PR #3, opened 2019-10-18
Create and Delete Profiles

cassiascheffer:
This PR adds a class to interface with the `/profiles` endpoint. It implements both `create` and `delete` actions.

See my review below for more details.


README.md
```diff
@@ -48,6 +46,66 @@ client = Bambora::Client.new(
...
+      'number': '4030000Smith010001234',
+      'expiry_month': '12',
+      'expiry_year': '23',
+      'cvd': '123',
+    },
+  },
+)
+# => {
+#       code: 1,
+#       message: 'Hup...want...buy.',
```
cassiascheffer, line 79: Poor Hup got motion sickness: https://www.youtube.com/watch?v=Sqj8xY0_Q0s

lib/bambora/client.rb
```diff
@@ -10,10 +11,14 @@ def initialize(options = {})
...
   end
 
+  def_delegators :connection, :request
+
   # Summary: Payment profiles store confidential payment information. Transactions can be processed against profiles.
   # Docs: https://dev.na.bambora.com/docs/guides/payment_profiles/
   # Endpoint: https://api.na.bambora.com/v1/profiles
-  def profiles; end
+  def profile
+    @profile ||= Bambora::Profile.new(self)
```
cassiascheffer: Passing the `Client` here so that the same client instance can be used for profiles, payment, etc.

lib/bambora/json_request.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Bambora
+  class JSONRequest
```
cassiascheffer, line 4: This is shared code for all JSON APIs. We'll also need an `XMLRequest` for some of the endpoints.

README.md
```diff
@@ -48,6 +46,66 @@ client = Bambora::Client.new(
...
+#         cvd_id: 123,
+#       },
+#    }
+```
+
+## Delete a Profile
+
+```ruby
+client.profile.delete(1234)
+# => {
```
cassiascheffer, line 100: The return from these methods is the same as what the existing `beanstream-ruby` library gives us.

lib/bambora/json_request.rb
```diff
@@ -0,0 +1,33 @@
...
+    attr_reader :client
+
+    def initialize(client)
+      @client = client
+      @path = ''
+    end
+
+    def request(method:, path:, params: {}, body: {})
+      parse_response(
```
cassiascheffer: Do we need any error handling here? If there is a network Error, we'll get an informative error from Excon.

Based on the existing gem, is possible that we'll get an error response back from Bambora, but those errors are in JSON as well, so I think we should handle them in the application.

lib/bambora/client.rb
```diff
@@ -11,10 +12,14 @@ def initialize(options = {})
...
     end
 
+    def_delegators :connection, :request
+
     # Summary: Payment profiles store confidential payment information. Transactions can be processed against profiles.
     # Docs: https://dev.na.bambora.com/docs/guides/payment_profiles/
     # Endpoint: https://api.na.bambora.com/v1/profiles
-    def profiles; end
+    def profile
+      @profile ||= Bambora::Profile.new(self)
```
cassiascheffer: Passing the Client here so that the same client instance can be used for profiles, payment, etc.

cassiascheffer:
[Bambora API Wrapper: Profiles API [3]](https://trello.com/c/3jftnjLn/1233-bambora-api-wrapper-profiles-api-3)

lib/bambora/json_request.rb
bilogub: i guess if server error happens(not application which will form correct json with error as the body) you won't get to JSON response instead you'll get JSON parse exception and this is our responsibility to handle it. Possible solution is to check for response code first and if it signals about error wrap it in valid JSON.

lib/bambora/json_request.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    private
+
+    def parse_response(resp)
+      symbolize_keys(JSON.parse(resp.body))
+    end
+
+    def symbolize_keys(obj)
+      return obj unless obj.is_a?(Hash)
```
bilogub: what if you get a hash which holds a collection of hashes e.g. 
```ruby
obj = {
  'user' => {
    'accounts' => [
      {
        'number' => '123',
      },
      {
        'number' => '456',
      },
    ]
  }
}
def symbolize_keys(obj)
  return obj unless obj.is_a?(Hash)

  obj.each_with_object({}) do |(k, v), hash|
    hash[k.to_sym] = symbolize_keys(v)
    hash
  end
end

symbolize_keys(obj) => {:user=>{:accounts=>[{"number"=>"123"}, {"number"=>"456"}]}}
```

lib/bambora/profile.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Bambora
+  class Profile < JSONRequest
+    def initialize(client)
+      super
+
+      @path = '/v1/profiles'
```
bilogub: would probably start to break into api versions here and namespacing the classes into `Bambora::V1::` in case users would want to stay on old version if supported and someone would develop V2 if available.

Gemfile.lock
```diff
@@ -7,10 +7,15 @@ PATH
 GEM
   remote: https://rubygems.org/
   specs:
+    addressable (2.7.0)
```
harrylewis, line 10: Were these changes added manually? I don't see any changes made to the `Gemfile`.

lib/bambora/json_request.rb
bilogub: took from rails source code:
```ruby
def _deep_transform_keys_in_object(object, &block)
      case object
      when Hash
        object.each_with_object({}) do |(key, value), result|
          result[yield(key)] = _deep_transform_keys_in_object(value, &block)
        end
      when Array
        object.map { |e| _deep_transform_keys_in_object(e, &block) }
      else
        object
      end
    end

    _deep_transform_keys_in_object(obj){ |key| key.to_sym } => {:USER=>{:ACCOUNTS=>[{:NUMBER=>"123"}, {:NUMBER=>"456"}]}}
```
You could wrap that method into your method

README.md
```diff
@@ -48,6 +46,66 @@ client = Bambora::Client.new(
 )
 ```
 
+### Profiles
+
+#### Create a Profile
+
+```ruby
+client.profile.create(
```
harrylewis, line 64: Great example here!

README.md
```diff
@@ -48,6 +46,66 @@ client = Bambora::Client.new(
 )
 ```
 
+### Profiles
+
+#### Create a Profile
```
harrylewis, line 61: Link to the original API documentation for these operations?

README.md
```diff
@@ -48,6 +46,66 @@ client = Bambora::Client.new(
...
+#         amount: 0,
+#         cvd_id: 123,
+#       },
+#    }
+```
+
+## Delete a Profile
+
+```ruby
+client.profile.delete(1234)
```
harrylewis: Bambora remote customer codes contain letters, and are therefore strings. Use a string input instead for the example.

lib/bambora/json_request.rb
cassiascheffer: Nice, thanks @bilogub!

lib/bambora/profile.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Bambora
+  class Profile < JSONRequest
```
harrylewis: More of a semantic thing, I find it a bit confusing that `Profile` inherits from `JSONRequest`. I infer from this inheritance chain that `Profile` is a type of `JSONRequest`, when in reality it contains multiple methods that can make JSON requests. This seems to me to be more like a concern/module.

From the perspective of the caller though, when you are using this client, you are only ever using one API method at a time, so from the point of view of the caller, it does only represent one JSON request.

Now I've just argued with myself, haha.

lib/bambora/profile.rb
```diff
@@ -0,0 +1,19 @@
...
+      super
+
+      @path = '/v1/profiles'
+    end
+
+    def create(card_data)
+      request(method: :post, path: @path, body: card_data)
+    end
+
+    def delete(remote_id)
```
harrylewis: To better align with the Bambora terminology, call this parameter, `customer_code`. Also, could this be a named parameter so that is clear? When looking at the documentation, I had to go to the implementation to determine what the input was. A named parameter would clear up that confusion immediately.

harrylewis (commented):
Love watching this shape up, and I'm learning a lot from it. I had a few comments to consider.

lib/bambora/profile.rb
cassiascheffer: I was going to call this `Domain` since it represents the Bambora domain, but then I realized that the domain also uses XML. I could perhaps name this `JSONDomain` and `XMLDomain`.

Gemfile.lock
cassiascheffer: There are different ways of managing dependencies for gems. The "bundler way" is to list all dependencies in the gemspec then call `gemspec` in the Gemfile to load dependencies listed there.

https://bundler.io/v2.0/guides/creating_gem.html#testing-our-gem

I don't know much about this project, but [hoe](https://www.zenspider.com/projects/hoe.html) has another way of managing dependencies.

spec/bambora/client_spec.rb
```diff
@@ -4,8 +4,21 @@
...
+
+        it 'returns a Bambora::V1::Profile' do
+          expect(subject.profile).to be_a Bambora::V1::Profile
+        end
+      end
+
+      context 'with a different version' do
+        subject { Bambora::Client.new(version: 'V2') }
+
+        it 'returns a raises an error' do
```
bilogub, line 19: is something missing here(body/pending + rewording)?

README.md
```diff
@@ -14,11 +15,16 @@ gem 'bambora-client'
...
+```bash
+bundle
+```
 
 Or install it yourself as:
 
-    $ gem install bambora-client
+```bash
+gem install bambora-client
+
```
harrylewis: Is the extra space needed? Not a blocker for this PR. This could be fixed when some Markdown linting is included 👀.

README.md
```diff
@@ -50,20 +54,89 @@ client = Bambora::Client.new(
 )
 ```
 
+### Profiles
+
+*Summary*: Payment profiles store confidential payment information. Transactions can be processed against profiles.
```
harrylewis, line 58: Thanks for cutting the line-length here.

lib/bambora/json_request.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    private
+
+    def parse_response(resp)
+      deep_transform_keys_in_object(JSON.parse(resp.body), &:to_sym)
+    end
+
+    def deep_transform_keys_in_object(object, &block)
+      case object
```
harrylewis, line 26: 👀

spec/bambora/client_spec.rb
harrylewis, line 19: Is this supposed to be empty?

harrylewis (approved):
Multi-threading here we come!

lib/bambora/json_request.rb
```diff
@@ -0,0 +1,38 @@
...
+    attr_reader :client
+
+    def initialize(client)
+      @client = client
+      @path = ''
+    end
+
+    def request(method:, path:, params: {}, body: {})
+      resp = client.request(method: method, path: path, params: params, body: body.to_json.to_s)
+      return parse_response(resp) if resp.status <= 300
```
bilogub: i see here a problem - what if it is 404. As a json client i want to get a response with 404 and empty collection.
instead i get a generic error which i have to parse(raw text) to interpret it something meaningful in my app.
https://jsonplaceholder.typicode.com/posts/1 - this returns me an object and this returns me an empty object
https://jsonplaceholder.typicode.com/posts/12123123123 with a status in the response header.

spec/bambora/client_spec.rb
cassiascheffer: Oof, sorry! Just pushed this!

lib/bambora/json_request.rb
cassiascheffer: @bilogub Do you think we should parse all 4xx level errors and only raise an erro on 5xx and 3xx?

bilogub: @cassidycodes i think i gave a falsey option in my prev comment regading possible solution here and misled you, sorry.
In nutshell we should relay anything we get from api back to our client app(as long as we don't have our own API doc to rely on - we are only sitting in the middle between API and client app passing bytes there and back) - that said parse response in begin/rescue block and pass it up the chain. In case we end up in rescue block compose 50x server response in json format. What do you think?

cassiascheffer: Sounds good! I just pushed a change that passes a hash with the error back to the user.

bilogub (approved):

PR #4, opened 2019-10-24
Ensure `Content-Type` header is set for JSON requests

harrylewis:
**Description**

This PR adds a few dependencies to `Bambora::Client`, and ensures that a `Content-Type` header of `application/json` is set for any request made from the descendants of `JSONRequest`.

Some clean up is also made to the documentation and specs. One of the notable changes is to update the `/profiles` API on `Bambora::Client` from `profile` to `profiles`. This makes it more RESTful, and better aligns with the rest of the API.

**What To Test**

+ [x] Can create a profile from `bin/console`
+ [x] Can delete a profile from `bin/console`

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ client = Bambora::Client.new(
 #### Create a Profile
 
 ```ruby
-client.profile.create(
+client.profiles.create(
   {
     language: 'en',
     comments: 'hello',
@@ -84,7 +84,7 @@ client.profile.create(
 ## Delete a Profile
 
 ```ruby
-client.profile.delete(customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A')
+client.profiles.delete(customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A')
 # => {
 #       :code => 1,
 #       :message => "Operation Successful"
--- a/README.md
+++ b/README.md
@@ -84,23 +84,11 @@ client.profile.create(
 ## Delete a Profile
 
 ```ruby
-client.profile.delete(customer_code: 'asdf1234')
+client.profile.delete(customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A')
 # => {
-#       code: 1,
-#       message: 'Hup...want...buy.',
-#       customer_code: 'asdf1234',
-#       validation: {
-#         id: '',
-#         approved: 1,
-#         message_id: 1,
-#         message: '',
-#         auth_code: '',
-#         trans_date: '',
-#         order_number: '',
-#         type: '',
-#         amount: 0,
-#         cvd_id: 123,
-#       },
+#       :code => 1,
+#       :message => "Operation Successful"
+#       :customer_code: "02355E2e58Bf488EAB4EaFAD7083dB6A",
 #    }
 ```
 
--- a/README.md
+++ b/README.md
@@ -63,33 +63,21 @@ client = Bambora::Client.new(
 ```ruby
 client.profile.create(
   {
-    'language': 'en',
-    'comments': 'hello',
-    'card':{
-      'name': 'Hup Podling',
-      'number': '4030000010001234',
-      'expiry_month': '12',
-      'expiry_year': '23',
-      'cvd': '123',
+    language: 'en',
+    comments: 'hello',
+    card: {
+      name: 'Hup Podling',
+      number: '4030000010001234',
+      expiry_month: '12',
+      expiry_year: '23',
+      cvd: '123',
     },
   },
 )
 # => {
-#       code: 1,
-#       message: 'Hup...want...buy.',
-#       customer_code: 'asdf1234',
-#       validation: {
-#         id: '',
-#         approved: 1,
-#         message_id: 1,
-#         message: '',
-#         auth_code: '',
-#         trans_date: '',
-#         order_number: '',
-#         type: '',
-#         amount: 0,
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -63,56 +63,32 @@ client = Bambora::Client.new(
...
-#         message_id: 1,
-#         message: '',
-#         auth_code: '',
-#         trans_date: '',
-#         order_number: '',
-#         type: '',
-#         amount: 0,
-#         cvd_id: 123,
-#       },
+#       :code => 1,
```
harrylewis, line 78: After creating a payment profile successfully, this is the response that I received back. It is identical to the one received when deleting a payment profile as well, down below.

lib/bambora/client.rb
```diff
@@ -1,9 +1,12 @@
 # frozen_string_literal: true
 
+require 'base64'
+require 'excon'
```
harrylewis, line 4: It sucks not having the Rails autoloader, and having to `require` everything explicitly.

lib/bambora/client.rb
```diff
@@ -46,7 +52,7 @@ def batchpayment_reports; end
     protected
 
     def connection
-      @connection ||= Excon.new(ENV.fetch('BAMBORA_API_URL'), headers: headers)
+      @connection ||= Excon.new(@base_url, headers: headers)
```
harrylewis, line 58: I know there was talk about allowing the usage of the sandbox URL, versus the production URL. I see allowing the former as a follow-up feature down the line. In any case, since this is a Bambora library, I think it makes sense to have the base URL hard-coded in here.

bin/console
```diff
@@ -8,5 +8,5 @@ require 'bambora'
 # with your gem easier. You can also use a different console, if you like.
 
 # (If you use this, don't forget to add pry to your Gemfile!)
-require "pry"
+require 'pry'
```
harrylewis, line 11: Only one linting error in the whole project to fix!

lib/bambora/json_request.rb
```diff
@@ -10,7 +10,14 @@ def initialize(client)
     end
 
     def request(method:, path:, params: {}, body: {})
-      resp = client.request(method: method, path: path, params: params, body: body.to_json.to_s)
+      resp = client.request(
+        method: method,
+        path: path,
+        params: params,
+        body: body.to_json.to_s,
+        headers: { 'Content-Type' => 'application/json' },
```
harrylewis, line 18: I am assuming since this class is called `JSONRequest`, that all requests should have this `Content-Type` header. It is required by Bambora's API.

spec/bambora/v1/profile_spec.rb
```diff
@@ -8,33 +8,25 @@ module V1
...
-            approved: 1,
-            message_id: 1,
-            message: '',
-            auth_code: '',
-            trans_date: '',
-            order_number: '',
-            type: '',
-            amount: 0,
-            cvd_id: 123,
-          },
```
harrylewis, line 29: Modified the tests to align with the new implementation.

README.md
```diff
@@ -61,58 +61,34 @@ client = Bambora::Client.new(
 #### Create a Profile
 
 ```ruby
-client.profile.create(
+client.profiles.create(
```
harrylewis, line 64: I updated the API to be more RESTful.

README.md
```diff
@@ -61,58 +61,34 @@ client = Bambora::Client.new(
...
+#       :message => "Operation Successful"
+#       :customer_code: "02355E2e58Bf488EAB4EaFAD7083dB6A",
 #    }
 ```
 
 ## Delete a Profile
 
 ```ruby
-client.profile.delete(customer_code: 'asdf1234')
+client.profiles.delete(customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A')
```
harrylewis, line 87: This is a real customer code, which no longer exists.

lib/bambora/client.rb
```diff
@@ -1,9 +1,15 @@
 # frozen_string_literal: true
 
+require 'base64'
+require 'excon'
+
+require 'bambora/headers'
+require 'bambora/v1/profile'
```
harrylewis, line 7: I definitely take autoloading for granted. I don't really understand how it works. Testing the gem with `bin/console` works without these `require`s. However, when I tried to integrate the gem into the HiMama repo., I was receiving messages like `Uninitialized Constant Bambora::V1`, which were solved in the context of Rails by including these explicitly.

I looked through the `paper_trail` gem, to see if this architecture was common, and it seems like this is correct.

My knowledge of including files in Ruby is limited, so I would love some insight here.

lib/bambora/v1/profile.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'bambora/json_request'
```
harrylewis, line 3: Needed an explicit `require` for Rails.

harrylewis (commented):
Self review.

lib/bambora/json_request.rb
bilogub: as a side note - might need to have default headers hash as an input merged with `Content-type`

harrylewis:
[Implement Bambora::Client.profiles on hm-proj [2]](https://trello.com/c/3NeViKu9/1284-implement-bamboraclientprofiles-on-hm-proj-1)

bilogub (approved):
Looks ok to me

PR #5, opened 2019-10-24
Restructure the library

harrylewis:
**Description**

This PR restructures the library architecture, with a few key ideas in mind.

+ `Bambora::Client` is the lowest available class, responsible for executing a network request, and must be explicitly instantiated (it now accepts and requires a `base_url` parameter)
+ `Bambora::JSONClient` inherits from `Bambora::Client`, with the `Content-Type` set and some convenience methods for parsing the response
+ A client from above is passed to a `*Resource`, to be used with specific actions that are allowed on the instance to make network requests

**What To Test**

+ [x] Can create a profile
+ [x] Can delete a profile
+ [x] Documentation and architecture make sense


README.md
```diff
@@ -31,13 +31,15 @@ You can initilize the client by passing a hash to it, or by passing a block.
 
 ```ruby
 client = Bambora::Client.new(
+  base_url: ENV.fetch('BAMBORA_BASE_URL') # Sandbox or Production URL
```
harrylewis, line 34: Now a `base_url` is required, and is passed explicitly so that the user can choose either production or sandbox.

README.md
```diff
@@ -47,72 +49,58 @@ You can also initilze a client with a sub-merchant-id:
 
 ```ruby
 client = Bambora::Client.new(
-  api_key: ENV.fetch('BAMBORA_API_KEY'),
-  merchant_id: ENV.fetch('BAMBORA_MERCHANT_ID'),
+  ...
```
harrylewis, line 52: DRY this up.

README.md
```diff
@@ -47,72 +49,58 @@ You can also initilze a client with a sub-merchant-id:
...
 ### Profiles
 
 *Summary*: Payment profiles store confidential payment information. Transactions can be processed against profiles.
 *Bambora Docs*: <https://dev.na.bambora.com/docs/guides/payment_profiles/>
 
+To use the profiles API, you must create an instance of the `Bambora::V1::ProfileResource` class.
+
+```ruby
+client = Bambora::JSONClient.new(...)
+profiles = Bambora::V1::ProfileResource.new(client: client, sub_path: '/v1/profiles')
```
harrylewis, line 68: This flow applies to all actions for profiles, therefore, brought this up to the top.

lib/bambora/client.rb
```diff
@@ -1,14 +1,16 @@
 # frozen_string_literal: true
 
+require 'base64'
+require 'excon'
```
harrylewis, line 4: These are required explicitly in order to be used in the client.

lib/bambora/client.rb
```diff
@@ -19,34 +21,10 @@ def initialize(options = {})
 
     def_delegators :connection, :request
 
-    def profile
```
harrylewis, line 22: Moved all of these methods to indiviual `*Resource` classes.

lib/bambora/json_client.rb
```diff
@@ -1,16 +1,16 @@
 # frozen_string_literal: true
 
 module Bambora
-  class JSONRequest
-    attr_reader :client
+  class JSONClient < Bambora::Client
```
harrylewis, line 4: Now, inherit from `Bambora::Client`. The difference here is the explicit `Content-Type` header below, and the JSON specific parsing methods.

lib/bambora/v1/batch_payment_report_resource.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Bambora
+  module V1
+    class BatchPaymentReportResource
```
harrylewis, line 5: Basic blueprint for each unimplemented API.

lib/bambora/v1/profile.rb
```diff
@@ -1,21 +0,0 @@
-# frozen_string_literal: true
-
-module Bambora
```
harrylewis, line 3: Renamed this to `ProfileResource`.

lib/bambora/v1/profile_resource.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Bambora
+  module V1
+    class ProfileResource
+      def initialize(client:, sub_path:)
+        @client = client
+        @sub_path = sub_path
```
harrylewis, line 8: Simply pass in the client and the sub-path. The reason the sub-path is needed is because when instantiating an instance of `Excon`, whatever is passed in as the base URL will get cut off after the last domain (e.g. `.com`).

Therefore,

```ruby
Excon.new('https://bambora.com/v1/profiles')
```

and

```ruby
Excon.new('https://bambora.com')
```

are equivalent. All requests will be made to `https://bambora.com`, and the remaining path needs to be passed in when making a request against the instance.

harrylewis (commented):
Self review.

README.md
```diff
@@ -47,72 +49,58 @@ You can also initilze a client with a sub-merchant-id:
...
+      name: 'Hup Podling',
+      number: '4030000010001234',
+      expiry_month: '12',
+      expiry_year: '23',
+      cvd: '123',
     },
   },
 )
 # => {
-#       code: 1,
```
cassiascheffer, line 78: Does the actual response differ from their docs, then? Or did I miss-read the docs?

lib/bambora/client.rb
cassiascheffer: I'm curious about this. When I tested the client myself, it worked without these lines. I may have run `rake build` when testing.

cassiascheffer (commented):
This is good! I like how this is evolving. 

I just left one question about the explicit `require`s in the client. This seems like something we shouldn't need. Or, at least, the `base64` one seems the most out of place.

README.md
harrylewis: Which documentation were you reading from?

harrylewis: @cassidycodes which documentation were you reading from?

cassiascheffer: https://dev.na.bambora.com/docs/references/payment_APIs/v1-0-5/

You have to click around to get to the `response`.

image

harrylewis: Right, I see. When I clicked through that I found under the `Profiles` accordion tab the response should look like this:

image

Which does align with the actual response I received.

lib/bambora/client.rb
harrylewis: Yes it also seemed to work for me before, but I can't remember how I got it to work. `base64` is not part of the core library, and there is a call to `Base64` at the bottom of this file.

README.md
cassiascheffer: Oooh, right. I got my alliterative endpoint names confused.

cassiascheffer (approved):
Thank you for doing this! 🎉

README.md
bilogub: i was saying you could have a higher level (JSON-)XML*Resource e.g.
```ruby
module Bambora
  module V1
    class JSONProfileResource < ProfileResource
      def initialize(sub_path:)
        super(client: Bambora::JSONClient.new(...), sub_path: sub_path)
      end
    end
  end
end
```
At this point you would build your user public api of "separate bricks"

bilogub: Less boilerplate but if current works for you - im ok with it as well.

bilogub (approved):
1 minor comment, just for thoughts

README.md
harrylewis: Yes I agree with this. I contemplated doing this, and decided to just leave it as is, and build these higher-level classes later, in order to narrow the scope of this PR.

harrylewis:
[Refactor bambora-client gem architecture [2]](https://trello.com/c/3BrvUYrn/1352-refactor-bambora-client-gem-architecture-2)

README.md
```diff
@@ -31,13 +31,15 @@ You can initilize the client by passing a hash to it, or by passing a block.
...
 client = Bambora::Client.new(
+  base_url: ENV.fetch('BAMBORA_BASE_URL') # Sandbox or Production URL
   api_key: ENV.fetch('BAMBORA_API_KEY'),
   merchant_id: ENV.fetch('BAMBORA_MERCHANT_ID'),
 )
 ```
 
 ```ruby
+  base_url: ENV.fetch('BAMBORA_BASE_URL') # Sandbox or Production URL
```
erawk, line 42: This should be `c.base_url` when using block-style instantiation.

README.md
erawk: Should the `*Resource` already know the `sub_path` for it's intended resource? Following the [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), I'd want to pass along as little as possible and have this library handle as much as possible for me.

lib/bambora/v1/profile_resource.rb
erawk: While this is an implementation short-coming of `Excon`, I don't see why the developer ergonomics require passing in the `sub_path`. Wouldn't it always be consistent for all profiles, and therefore be an implementation detail of the class, rather than developer knowledge passed in?

erawk (commented):
I know you already have merged this in, but if there are further edits coming, please consider some of the developer experience recommendations I've added.

PR #6, opened 2019-10-28
Implement initial Payments API

harrylewis:
**Description**

This PR implements an initial set of methods that can be used to access the Bambora Payments API.

**What To Test**

+ [x] Can make a credit card payment
+ [x] Can make a payment profile payment

*Test cards can be found [here](https://dev.na.bambora.com/docs/references/payment_APIs/test_cards/).*

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -57,6 +57,11 @@ Metrics/MethodLength:
   CountComments: false
   Max: 100
 
+Metrics/ModuleLength:
+  # These use file-spanning module blocks, making this cop useless
+  Exclude:
+    - 'spec/**/*'
+
 Metrics/PerceivedComplexity:
   Max: 50
 
--- a/README.md
+++ b/README.md
@@ -87,9 +87,9 @@ profiles.create(
   },
 )
 # => {
-#       :code => 1,
-#       :message => "Operation Successful",
-#       :customer_code => "02355E2e58Bf488EAB4EaFAD7083dB6A",
+#      :code => 1,
+#      :message => "Operation Successful",
+#      :customer_code => "02355E2e58Bf488EAB4EaFAD7083dB6A",
 #    }
 ```
 
@@ -98,9 +98,9 @@ profiles.create(
 ```ruby
 profiles.delete(customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A')
 # => {
-#       :code => 1,
-#       :message => "Operation Successful",
-#       :customer_code => "02355E2e58Bf488EAB4EaFAD7083dB6A",
+#      :code => 1,
+#      :message => "Operation Successful",
+#      :customer_code => "02355E2e58Bf488EAB4EaFAD7083dB6A",
 #    }
 ```
 
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ To use the payments API, you must create an instance of the `Bambora::V1::Paymen
 
 ```ruby
 client = Bambora::JSONClient.new(...)
-payments = Bambora::V1::PaymentResource.new(client: client, sub_path: '/v1/payments')
+payments = Bambora::V1::PaymentResource.new(client: client)
 ```
 
 Once the resource instance has been instantiated, actions can be made against the API.
--- a/README.md
+++ b/README.md
@@ -104,6 +104,86 @@ profiles.delete(customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A')
 #    }
 ```
 
+### Payments
+
+*Summary*: Process payments using Credit Card, Payment Profile, Legato Token, Cash, Cheque, Interac, Apple Pay, or
+Android Pay.
+*Bambora Docs*: <https://dev.na.bambora.com/docs/references/payment_APIs/>
+
+To use the payments API, you must create an instance of the `Bambora::V1::PaymentResource` class.
+
+```ruby
+client = Bambora::JSONClient.new(...)
+payments = Bambora::V1::PaymentResource.new(client: client, sub_path: '/v1/payments')
+```
+
+Once the resource instance has been instantiated, actions can be made against the API.
+
+#### Make A Payment
+
+This is a lower level method that can be used to make a payment using any method. Below are some more specific and
+high level convenience methods.
+
+```ruby
+payments.make_payment(
+  {
+    amount: 50,
+    payment_method: 'card',
+    card: {
+      name: 'Hup Podling',
+      number: '4504481742333',
+      expiry_month: '12',
+      expiry_year: '20',
+      cvd: '123',
+    },
+  },
+)
+# => {
+       :id => "10000000",
+       :authorizing_merchant_id => 300000000,
+       :approved => "1",
+       :message_id => "1",
+       :message => "Approved",
+       :auth_code => "TEST",
+       :created => "2019-10-28T07:12:11",
+       :order_number => "10000000",
+       :type => "P",
+       :payment_method => "CC",
+       :risk_score => 0.0,
+       :amount => 50.0,
+       :custom => { :ref1 => "", :ref2 => "", :ref3 => "", :ref4 => "", :ref5 => "" },
+       :card =>
+         {
+           :card_type => "VI",
+           :last_fouri => "2333",
+           :address_match => 0,
+           :postal_result => 0,
+           :avs_result => "0",
+           :cvd_result => "1",
+           :avs => { :id => "U", :message => "Address information is unavailable.", :processed => false }
+         },
+       :links =>
+         [
+           { :rel => "void", :href => "https://api.na.bambora.com/v1/payments/10000000/void", :method => "POST" },
+           { :rel => "return", :href => "https://api.na.bambora.com/v1/payments/10000000/returns", :method => "POST" }
+         ]
+     }
+```
+
+##### Make A Payment With A Payment Profile
+
+```ruby
+payments.make_payment_with_payment_profile(customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A', amount: 50)
+# => {
+       :id => "10000000",
+       :authorizing_merchant_id => 300000000,
+       :approved => "1",
+       :message_id => "1",
+       :message => "Approved",
+       ...
+     }
+```
+
 ## Development
 
 After checking out the repo, run `bin/setup` to install dependencies. Then, run `rake spec` to run the tests. You can
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ client = Bambora::Client.new(
 
 ```ruby
 client = Bambora::Client.new do |c|
-  base_url: ENV.fetch('BAMBORA_BASE_URL') # Sandbox or Production URL
+  c.base_url: ENV.fetch('BAMBORA_BASE_URL') # Sandbox or Production URL
   c.api_key = ENV.fetch('BAMBORA_API_KEY')
   c.merchant_id = ENV.fetch('BAMBORA_MERCHANT_ID')
 end
--- a/README.md
+++ b/README.md
@@ -139,35 +139,35 @@ payments.make_payment(
   },
 )
 # => {
-       :id => "10000000",
-       :authorizing_merchant_id => 300000000,
-       :approved => "1",
-       :message_id => "1",
-       :message => "Approved",
-       :auth_code => "TEST",
-       :created => "2019-10-28T07:12:11",
-       :order_number => "10000000",
-       :type => "P",
-       :payment_method => "CC",
-       :risk_score => 0.0,
-       :amount => 50.0,
-       :custom => { :ref1 => "", :ref2 => "", :ref3 => "", :ref4 => "", :ref5 => "" },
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -87,9 +87,9 @@ profiles.create(
   },
 )
 # => {
-#       :code => 1,
-#       :message => "Operation Successful",
-#       :customer_code => "02355E2e58Bf488EAB4EaFAD7083dB6A",
+#      :code => 1,
```
harrylewis, line 90: Fixed spacing/alignment.

README.md
```diff
@@ -39,7 +39,7 @@ client = Bambora::Client.new(
 
 ```ruby
-  base_url: ENV.fetch('BAMBORA_BASE_URL') # Sandbox or Production URL
+  c.base_url: ENV.fetch('BAMBORA_BASE_URL') # Sandbox or Production URL
```
harrylewis, line 42: Fixed a missing object reference in the example.

README.md
```diff
@@ -98,12 +98,92 @@ profiles.create(
...
+high level convenience methods.
+
+```ruby
+payments.make_payment(
+  {
+    amount: 50,
+    payment_method: 'card',
+    card: {
+      name: 'Hup Podling',
+      number: '4504481742333',
```
harrylewis, line 134: I used some [test credit card](https://dev.na.bambora.com/docs/references/payment_APIs/test_cards/), which allow us to make test payments.

lib/bambora/v1/payment_resource.rb
```diff
@@ -3,15 +3,24 @@
 module Bambora
   module V1
     class PaymentResource
-      # Summary: Create and modify payments.
-      # Note: The link below links to all apis includding profiles and tokenization. There aren't great docs explaining
-      #       the /payments endpoints alone.
-      # Docs: https://dev.na.bambora.com/docs/references/payment_APIs/
-      #       https://dev.na.bambora.com/docs/references/payment_SDKs/take_payments/?shell#
-      # Endpoint: https://api.na.bambora.com/v1/payments
```
harrylewis, line 11: Moved reference to some of these links to the `README` documentation.

lib/bambora/v1/payment_resource.rb
```diff
@@ -3,15 +3,24 @@
...
-      def initialize(client:, sub_path:)
+      def initialize(client:)
         @client = client
-        @sub_path = sub_path
+        @sub_path = '/v1/payments'
```
harrylewis, line 14: A change that has been queued up for the `Bambora::V1::ProfileResource`. Remove the required `sub_path` initialization variable, as the endpoint is known beforehand, and can therefore be a class-level detail.

lib/bambora/v1/payment_resource.rb
```diff
@@ -3,15 +3,24 @@
...
+      end
+
+      def make_payment(payment_data)
+        @client.request(method: :post, path: @sub_path, body: payment_data)
```
harrylewis, line 18: Lower-Level API method.

lib/bambora/v1/payment_resource.rb
```diff
@@ -3,15 +3,24 @@
...
+      end
+
+      def make_payment_with_payment_profile(customer_code:, amount:)
+        make_payment(
```
harrylewis, line 22: Convenience method for our bespoke use case. Others can be created later as well.

harrylewis (commented):
Self review.

lib/bambora/v1/payment_resource.rb
erawk: Hm, while I think it is important to have it in the `README`, the plus of having this here is with `pry-doc` commands. While navigating the code in a debugger, having these additional doc snippets available would really help.

lib/bambora/v1/payment_resource.rb
```diff
@@ -3,15 +3,24 @@
...
+          amount: amount,
+          payment_method: 'payment_profile',
+          payment_profile: {
+            customer_code: customer_code,
+            card_id: 1,
```
erawk: Is this always going to `1`? What does `1` mean?

erawk (commented):
Looks great. Couple minor queries.

harrylewis:
[Bambora API Wrapper: Payments API [3]](https://trello.com/c/NwquxvdC/1234-bambora-api-wrapper-payments-api-3)

README.md
```diff
@@ -98,9 +98,89 @@ profiles.create(
...
+
+*Summary*: Process payments using Credit Card, Payment Profile, Legato Token, Cash, Cheque, Interac, Apple Pay, or
+Android Pay.
+*Bambora Docs*: <https://dev.na.bambora.com/docs/references/payment_APIs/>
+
+To use the payments API, you must create an instance of the `Bambora::V1::PaymentResource` class.
+
+```ruby
+client = Bambora::JSONClient.new(...)
+payments = Bambora::V1::PaymentResource.new(client: client, sub_path: '/v1/payments')
```
bilogub: `sub_path` 🤔

harrylewis: @bilogub good catch, will update.

cassiascheffer (commented):
Seconding Ed's comments here!

lib/bambora/v1/payment_resource.rb
harrylewis: This is a parameter that we have included in our own codebase. Within a Bambora payment profile, it is possible to store multiple cards. Specifying the card ID allows you to use one of multiple cards on the same profile. Since we do not currently use this functionality, this parameter should always be set to `1`.

I realize this looks like a magic number, so I have extracted it to a class-level constant.

.rubocop.yml
```diff
@@ -57,6 +57,11 @@ Metrics/MethodLength:
   CountComments: false
   Max: 100
 
+Metrics/ModuleLength:
```
harrylewis, line 60: Specifically ignored the `Metrics/ModuleLength` to prevent these errors from continuing to crop up. Test blocks can become intentionally long.

README.md
```diff
@@ -98,9 +98,89 @@ profiles.create(
...
+
+*Summary*: Process payments using Credit Card, Payment Profile, Legato Token, Cash, Cheque, Interac, Apple Pay, or
+Android Pay.
+*Bambora Docs*: <https://dev.na.bambora.com/docs/references/payment_APIs/>
+
+To use the payments API, you must create an instance of the `Bambora::V1::PaymentResource` class.
+
+```ruby
+client = Bambora::JSONClient.new(...)
+payments = Bambora::V1::PaymentResource.new(client: client)
```
harrylewis, line 117: Removed the `sub_path` initialization variable.

lib/bambora/v1/payment_resource.rb
harrylewis: Added this back!

lib/bambora/v1/payment_resource.rb
```diff
@@ -3,15 +3,32 @@
...
+        @client.request(method: :post, path: @sub_path, body: payment_data)
+      end
+
+      def make_payment_with_payment_profile(customer_code:, amount:)
+        make_payment(
+          amount: amount,
+          payment_method: 'payment_profile',
+          payment_profile: {
+            customer_code: customer_code,
+            card_id: CREDIT_CARD_ID_DEFAULT,
```
harrylewis: Now using the `CREDIT_CARD_ID_DEFAULT`, rather than a magic number. Do you think I allow this to be passed in to the method as a parameter. My only issue with this, is that if we allow this, where do we draw the line in terms of what we do and don't allow to be passed in.

This method is meant to be more bespoke for our use case, however I do realize that given this is now OSS, it should be more generalized. If much more customization is needed, I would argue that using the lower level `#make_payment` method would be more useful, than having this method take on many many parameters. Unless we had a general options hash parameter, which could be merged in with the default data.

I'm okay to do something like `CREDIT_CARD_ID_DEFAULT || card_id`, with `card_id` being a `nil` default parameter in the method signature, if you think this would be the best way forward.

lib/bambora/v1/payment_resource.rb
cassiascheffer: Since this is a public gem though, shouldn't we allow users to pass `card_id` into the method?

harrylewis (commented):
Thank you for all the feedback!

lib/bambora/v1/payment_resource.rb
cassiascheffer: I think we could pass `card_id` in with a default so that this method can be applied to more broad use-cases. I'm wary of having a method that makes assumptions based on our specific needs. What if those needs change?

> My only issue with this, is that if we allow this, where do we draw the line in terms of what we do and don't allow to be passed in

What else is missing? The method name makes it clear that we're using a `payment_profile`, but it does not make it clear that we're using `card_id = 1`. I think that convenience methods like this are good if they are clear about what they are doing.

harrylewis: @cassidycodes in this case, [according to the documentation](https://dev.na.bambora.com/docs/references/payment_APIs/v1-0-5/), when making a payment using a payment profile, there are 3 pieces of data that can be passed, of which 2 are required:

+ Customer Code (required)
+ Card ID (required)
+ Complete

So it does look like the Card ID is required. In our use case, we always keep it at `1`, because we currently do not allow multiple credit cards. I will make this a named argument with a default value of `1` like you said.

The Complete field indicates whether the payment should be completed or not. If it is not completed, the payment is made as a pre-authorization, meaning it will put a hold of the amount on the payment on the account, which will be subsequently released after a few days.

Since this is the only other field, I will also implement this as a named parameter. To prevent unintended charges, I will default this value to `false`.

lib/bambora/v1/payment_resource.rb
```diff
@@ -9,9 +9,25 @@ class PaymentResource
...
         @client = client
-        @sub_path = sub_path
+        @sub_path = '/v1/payments'
+      end
+
+      def make_payment(payment_data)
+        @client.request(method: :post, path: @sub_path, body: payment_data)
+      end
+
+      def make_payment_with_payment_profile(customer_code:, amount:, card_id: 1, complete: false)
```
cassiascheffer, line 21: Hey, I know we've gone back and forth on this a lot, but this looks like a magic number again. Perhaps some documentation above this method would help clarify the usage here. Perhaps we can do that in a follow up PR.

cassiascheffer (approved):

lib/bambora/v1/payment_resource.rb
cassiascheffer: Ah, sorry, I didn't see this comment because of how GitHub links to changes from the notifications tab. Blargh. This all sounds good to me!

erawk: Keep it specific to our use-case. Should we or others need to generalize this, additional PRs can be created.

lib/bambora/v1/payment_resource.rb
```diff
@@ -3,15 +3,32 @@
 module Bambora
   module V1
     class PaymentResource
+      CREDIT_CARD_ID_DEFAULT = 1
```
erawk: Is this `1` because a payment profile could have multiple credit cards associated with it but in practice never does for our use case?

lib/bambora/v1/payment_resource.rb
erawk: How does this handle failures? Failure on the network? Failure with the transaction?

erawk (commented):
Looks good. Just a couple clarifying questions.

lib/bambora/v1/payment_resource.rb
harrylewis: Yes exactly.

lib/bambora/v1/payment_resource.rb
harrylewis: @erawk handling failures like this should be happening at the client level (`Bambora::JSONClient`). You can see that we do [*some* error-handling](https://github.com/HiMamaInc/bambora-client/blob/master/lib/bambora/json_client.rb#L15), but this could be more robust and `rescue` from more errors, such as ones that would result from the issues you mentioned above.

Comparing the error handling done in our gem versus the `beanstream` gem, it is possible to see that there is in fact [error handling for some of the errors you mentioned](https://github.com/bambora-na/beanstream-ruby/blob/master/lib/beanstream/transaction.rb#L87). I think this work is necessary moving forward, but I think is out of scope of this PR. I have created a [tile](https://trello.com/c/teSIlyMz) to address this.

As a safety precaution I searched BugSnag for some of the text that was contained in the `beanstream` gem REST client error raising I linked to above, and could not find a single record that any of these errors had occurred. To an end-user, I imagine this would appear as a 500 error, at which point they would have to go back, and re-input the action.

Given the low probability of this happening, given my searching through BugSnag, I feel it is okay to push forward, but prioritize this failure handling.

erawk: Yes, follow up tile makes sense. I think the client should handle some category of errors, like network, server communication and TLS issues. Others ought to be handled by the consumer of the client as you suggest.

PR #9, opened 2019-11-07
Renaming make_payment to create & adding alias

cassiascheffer:
I renamed `make_payment` to `create` as it is more technically
correct when describing a rest API. However, `make_payment` does sound
more correct in an English sentence. I have added `make_payment` as an
alias of `create`.

I've made the same changes for `make_payment_with_payment_profile`.`

### DIFF
--- a/lib/bambora/v1/payment_resource.rb
+++ b/lib/bambora/v1/payment_resource.rb
@@ -46,7 +46,6 @@ module Bambora
         @client.request(method: :post, path: @sub_path, body: payment_data)
       end
 
-      # An alias of +create+.
       alias make_payment create
 
       # Make a payment with a credit card. Aliased as +make_payment_with_payment_profile+.
@@ -60,7 +59,12 @@ module Bambora
       #     customer_code: '2355E2e58Bf488EAB4EaFAD7083dB6A', amount: 50, complete: false
       #   )
       #
-      # @param payment_data [Hash] All information relevant to making a payment.
+      # @param customer_code [String] Bambora's payment profile ID.
+      # @param amount [Float] A decimal value in dollars. Uses up to two decimal places. Max value is account specific.
+      #   Default max value is 1000.
+      # @param card_id [Integer] Default +1+. Which credit card to use. Starts at 1 for the first card. You must
+      #   configure how many cards can be stored by visiting the profile options in the back office.
+      # @param complete [Boolean] Default +false+. Set to false for Pre-Authorize, and true to complete a payment.
       #
       # @return [Hash] Indicating success or failure of the operation.
       def create_with_payment_profile(customer_code:, amount:, card_id: 1, complete: false)
@@ -75,7 +79,6 @@ module Bambora
         )
       end
 
-      # An alias of =create_with_payment_profile+.
       alias make_payment_with_payment_profile create_with_payment_profile
     end
   end
--- a/lib/bambora/v1/payment_resource.rb
+++ b/lib/bambora/v1/payment_resource.rb
@@ -14,12 +14,57 @@ module Bambora
         @sub_path = '/v1/payments'
       end
 
-      def make_payment(payment_data)
+      # Make a payment with a credit card. Also aliased as +make_payment+.
+      #
+      #
+      # @example
+      #
+      #   client = Bambora::JSONClient(base_url: '...', api_key: '...', merchant_id: '...')
+      #   payments = Bambora::V1::PaymentResource(client: client)
+      #   payments.create(
+      #     {
+      #       amount: 50,
+      #       payment_method: 'card',
+      #       card: {
+      #         name: 'Hup Podling',
+      #         number: '4504481742333',
+      #         expiry_month: '12',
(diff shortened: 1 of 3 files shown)


spec/bambora/v1/payment_resource_spec.rb
```diff
@@ -102,7 +102,15 @@ module V1
         end
       end
 
-      describe '#make_payment_with_payment_profile' do
+      describe '#make_payment' do
+        subject { Bambora::V1::PaymentResource.new(client: client) }
+
+        it 'is an alias of #create_with_payment_profile' do
+          expect(subject.method(:make_payment)).to eq(subject.method(:create))
```
harrylewis, line 109: Love this assertion, I didn't know `#method` existed.

lib/bambora/v1/payment_resource.rb
```diff
@@ -14,12 +14,57 @@ def initialize(client:)
...
+      #         cvd: '123',
+      #       },
+      #     },
+      #   )
+      #
+      # @param payment_data [Hash] All information relevant to making a payment.
+      #
+      # @see https://dev.na.bambora.com/docs/references/payment_APIs/
+      #
+      # @see https://dev.na.bambora.com/docs/references/payment_SDKs/take_payments/?shell#
```
harrylewis, line 42: Thanks for including these links here, for anything that we fail to describe.

lib/bambora/v1/payment_resource.rb
```diff
@@ -14,12 +14,57 @@ def initialize(client:)
...
+      #
+      # @return [Hash] Indicating success or failure of the operation.
+      def create(payment_data)
         @client.request(method: :post, path: @sub_path, body: payment_data)
       end
 
-      def make_payment_with_payment_profile(customer_code:, amount:, card_id: 1, complete: false)
-        make_payment(
+      # An alias of +create+.
```
harrylewis: Is the comment necessary? I feel that for anyone looking at the source code, this aliasing will already be clear.

lib/bambora/v1/payment_resource.rb
```diff
@@ -29,6 +74,9 @@ def make_payment_with_payment_profile(customer_code:, amount:, card_id: 1, compl
           },
         )
       end
+
+      # An alias of =create_with_payment_profile+.
```
harrylewis: If these alias comments remain, change the `=` to a `+`.

lib/bambora/v1/payment_resource.rb
```diff
@@ -14,12 +14,57 @@ def initialize(client:)
...
+      #
+      # @example
+      #
+      #   client = Bambora::JSONClient(base_url: '...', api_key: '...', merchant_id: '...')
+      #   payments = Bambora::V1::PaymentResource(client: client)
+      #   payments.create_with_payment_profile(
+      #     customer_code: '2355E2e58Bf488EAB4EaFAD7083dB6A', amount: 50, complete: false
+      #   )
+      #
+      # @param payment_data [Hash] All information relevant to making a payment.
```
harrylewis: Expand the `@param` definitions to explicitly name and explain all the named parameters?

harrylewis (commented):
Functionally this all looks great. I like the switch to prioritizing a RESTful interface. What you indicated in your PR description, the `make_payment` terminology reading more clearly in a sentence, I think is what drew me in to use that naming convention. I like the aliasing though, to give consumers of this gem the option to write their code this way.

I had some small comments mostly about the documentation. I'll approve once they are addressed.

safdarzeeshan (approved):
Thanks for making these updates to better follow restful standards. 
Though I can see why this naming is also tricky because when I read `create_with_payment_profile` it sounds like I am creating a payment profile. So having both options is helpful 👍

lib/bambora/v1/payment_resource.rb
cassiascheffer: It's not! I thought it was, but requesting docs on this method will return the docs for `create`. I'll remove it.

lib/bambora/v1/payment_resource.rb
```diff
@@ -14,12 +14,61 @@ def initialize(client:)
...
+      #   payments.create_with_payment_profile(
+      #     customer_code: '2355E2e58Bf488EAB4EaFAD7083dB6A', amount: 50, complete: false
+      #   )
+      #
+      # @param customer_code [String] Bambora's payment profile ID.
+      # @param amount [Float] A decimal value in dollars. Uses up to two decimal places. Max value is account specific.
+      #   Default max value is 1000.
+      # @param card_id [Integer] Default +1+. Which credit card to use. Starts at 1 for the first card. You must
+      #   configure how many cards can be stored by visiting the profile options in the back office.
+      # @param complete [Boolean] Default +false+. Set to false for Pre-Authorize, and true to complete a payment.
```
harrylewis, line 67: I can see you've really dug in to the Bambora documentation to find this information! Thanks for being so thorough.

harrylewis (approved):
Looks awesome, thanks for the documentation updates.

PR #10, opened 2019-11-08
Add Per-resource API Keys

cassiascheffer:
This library was originally designed to use one API key per client. This PR changes this behaviour so that we can use one API key per resource.

While doing this work, tests started to fail for Excon because it was ecpecting the request body to ba a string. Im not sure why we didn't see this behaviour before. I have added a helper method to stringify these request bodies.


bambora-client.gemspec
```diff
@@ -40,6 +40,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'excon', '~> 0.68.0'
   spec.add_development_dependency 'bundler', '~> 1.17'
   spec.add_development_dependency 'pry', '~> 0.12.0'
+  spec.add_development_dependency 'pry-byebug', '~> 3.7.0'
```
cassiascheffer, line 44: I wanted step-through debugging to figure out why Excon was throwing an error.

lib/bambora/headers.rb
```diff
@@ -3,11 +3,15 @@
...
     class << self
-      def build(passcode:, sub_merchant_id: nil)
-        headers = { 'Authorization' => "Passcode #{passcode}" }
+      def build(api_key:, merchant_id:, sub_merchant_id: nil)
+        headers = { 'Authorization' => "Passcode #{passcode(merchant_id, api_key)}" }
         headers['Sub-Merchant-Id'] = sub_merchant_id unless sub_merchant_id.nil?
         headers
       end
+
+      def passcode(merchant_id, api_key)
```
cassiascheffer, line 12: Moving passcode here since it's really a concern of `Headers`, not `Client`.

lib/bambora/v1/payment_resource.rb
```diff
@@ -3,14 +3,11 @@
 module Bambora
   module V1
     class PaymentResource
-      # Summary: Create and modify payments.
```
cassiascheffer, line 6: This now lives in the documentation on the methods below.

lib/bambora/json_client.rb
```diff
@@ -2,22 +2,61 @@
...
+    def request(method:, path:, body: {}, api_key:)
+      request_options = {
         method: method,
         path: path,
-        params: params,
-        body: body.to_json.to_s,
-        headers: { 'Content-Type' => 'application/json' }.merge(headers),
-      )
-
+        body: request_body(body),
```
cassiascheffer: I'm not sure why this stated happening, but Excon started to complain that the request body was not a string. I've added this method fro stringifying it.

cassiascheffer:
[Bambora::Client Allow for per-resource API keys](https://trello.com/c/lZnjzMeb/1454-bamboraclient-allow-for-per-resource-api-keys)

albertxie (approved):
Looks good to me!

lib/bambora/json_client.rb
```diff
@@ -2,22 +2,61 @@
 
 module Bambora
   class JSONClient < Bambora::Client
-    def request(method:, path: '', params: {}, body: {}, headers: {})
-      resp = connection.request(
+    # Make a JSON Request.
+    #
+    # @example
+    #
+    #   client = Bambora::JSONClient(base_url: '...', api_key: '...', merchant_id: '...')
```
harrylewis, line 9: Remove `api_key`.

README.md
```diff
@@ -32,15 +32,13 @@ You can initilize the client by passing a hash to it, or by passing a block.
...
 client = Bambora::Client.new(
   base_url: ENV.fetch('BAMBORA_BASE_URL') # Sandbox or Production URL
-  api_key: ENV.fetch('BAMBORA_API_KEY'),
   merchant_id: ENV.fetch('BAMBORA_MERCHANT_ID'),
 )
 ```
 
 ```ruby
   c.base_url: ENV.fetch('BAMBORA_BASE_URL') # Sandbox or Production URL
```
harrylewis: I think I made a mistake here on a previous commit. Can you change the `:` to a `=`?

spec/bambora/v1/payment_resource_spec.rb
```diff
@@ -103,7 +102,7 @@ module V1
       end
 
       describe '#make_payment' do
-        subject { Bambora::V1::PaymentResource.new(client: client) }
+        subject { Bambora::V1::PaymentResource.new(client: client, api_key: api_key) }
```
harrylewis: Be consistent about `subject` definitions by using `described_class`, as above? Maybe DRY up initialization by moving to the top of the spec file.

bambora-client.gemspec
```diff
@@ -38,8 +38,10 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
 
   spec.add_dependency 'excon', '~> 0.68.0'
+  spec.add_dependency 'faraday', '~> 0.17'
```
cassiascheffer, line 41: Added Faraday for a more clear HTTP interface.

lib/bambora/client.rb
```diff
@@ -19,20 +19,30 @@ def initialize(options = {})
...
+
+    def delete(path:, headers:)
+      connection.delete(path) do |req|
+        req.headers = headers
+      end
+    end
+
+    def connection
+      @connection ||= Faraday.new(url: base_url) do |f|
+        f.adapter :excon
```
cassiascheffer, line 39: Still using Excon under the hood.

lib/bambora/json_client.rb
```diff
@@ -2,24 +2,106 @@
 
 module Bambora
   class JSONClient < Bambora::Client
-    def request(method:, path: '', params: {}, body: {}, headers: {})
```
cassiascheffer, line 5: After causing a bug by accidentally removing `to_json.to_s` on ln 10, I chose to add faraday as an interface to Excon. It wasn't until I was reviewing this PR that the error I was seeing was actually my mistake 🤦‍♀ 

Adding Faraday did force me to make explicit methods for each request verb here, which I think makes the interface to each method more clear.

cassiascheffer: Ultimately, what inspired this change is that I needed one more arg for this method and rubocop got mad about the number of args. I considered using a hash instead, but the internals of the method got kind of messy when I did that.

lib/bambora/client.rb
```diff
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
 require 'base64'
-require 'excon'
+require 'faraday'
 
 module Bambora
   class Client
     extend Forwardable
```
harrylewis: Remove this, since the `def_delegators` directive was removed below.

lib/bambora/client.rb
```diff
@@ -19,20 +19,30 @@ def initialize(options = {})
...
-    def connection
-      @connection ||= Excon.new(base_url, headers: headers)
+    def get(path:, params:, headers:)
+      connection.get(path, params, headers)
     end
 
-    def headers
-      Bambora::Headers.build(passcode: passcode, sub_merchant_id: sub_merchant_id)
+    def post(path:, body:, headers:)
+      connection.post(path, body.to_json.to_s, headers)
```
harrylewis: Should the `.to_json.to_s` be moved to the `Bambora::JSONClient`?

lib/bambora/client.rb
```diff
@@ -19,20 +19,30 @@ def initialize(options = {})
...
+    end
+
+    def delete(path:, headers:)
+      connection.delete(path) do |req|
+        req.headers = headers
```
harrylewis, line 33: Was this done using a block so that you didn't have to unnecessarily set the `body` parameter in the call?

lib/bambora/json_client.rb
```diff
@@ -2,24 +2,108 @@
...
+    #
+    # @param method [Symbol] Indicating request verb.
+    # @param path [String] Indicating request path.
+    # @param body [Hash] Data to be sent in the body of the request.
+    # @param api_key [String] Indicating the API Key to be used with the request.
+    #
+    # @return [Hash] Indicating success or failure of the operation.
+    def post(path:, body:, api_key:)
+      parse_response(
+        super(path: path, body: body, headers: build_headers(api_key)),
```
harrylewis: Use `.to_json.to_s` here.

lib/bambora/json_client.rb
```diff
@@ -2,24 +2,108 @@
...
+      )
+    end
+
+    # Make a DELETE Request.
+    #
+    # @example
+    #
+    #   client = Bambora::JSONClient(base_url: '...', api_key: '...', merchant_id: '...')
+    #
+    #   client.delete(path: 'v1/profiles/asdf1234', api_key: '...')
```
harrylewis: Use `02355E2e58Bf488EAB4EaFAD7083dB6A` in the path for consistency.

lib/bambora/json_client.rb
```diff
@@ -2,24 +2,108 @@
...
+    def delete(path:, api_key:)
+      parse_response(
+        super(path: path, headers: build_headers(api_key)),
+      )
     end
 
     private
 
+    def build_headers(api_key)
+      { 'Content-Type' => 'application/json' }.merge(super(api_key))
```
harrylewis, line 100: I like how you have centralized this call.

lib/bambora/json_client.rb
harrylewis: I like the way this turned out. Much less configuration on our part to get various HTTP verbs to work correctly. `faraday` gives us what we need out of the box, and still uses `excon`. I think this is a great update and will help future us.

lib/bambora/v1/payment_resource.rb
```diff
@@ -3,14 +3,11 @@
...
-      # Note: The link below links to all apis includding profiles and tokenization. There aren't great docs explaining
-      #       the /payments endpoints alone.
-      # Docs: https://dev.na.bambora.com/docs/references/payment_APIs/
-      #       https://dev.na.bambora.com/docs/references/payment_SDKs/take_payments/?shell#
-      # Endpoint: https://api.na.bambora.com/v1/payments
-      def initialize(client:)
+      attr_reader :api_key, :client, :sub_path
```
harrylewis, line 6: Thanks for alphabetizing this.

lib/bambora/v1/payment_resource.rb
```diff
@@ -43,7 +40,7 @@ def initialize(client:)
       #
       # @return [Hash] Indicating success or failure of the operation.
       def create(payment_data)
-        @client.request(method: :post, path: @sub_path, body: payment_data)
+        client.post(path: sub_path, body: payment_data, api_key: api_key)
```
harrylewis, line 43: Look at the difference here:

```ruby
client.request(method: :post, ...)
```

versus

```ruby
client.post(...)
```

Much more terse and clear, I love it.

harrylewis (commented):
I like the use of `faraday` here to simplify the network calls. I had a few comments here to address, once they are done it should be good to approve!

lib/bambora/client.rb
cassiascheffer: Yeah, faraday uses positional arguments, so this would have been `connection.delete(path, {}, headers)`

spec/bambora/v1/payment_resource_spec.rb
cassiascheffer: I can do that. But note that my only change here was to add the `api_key` to make this test pass with the new syntax.

harrylewis (approved):
Let's keep this moving!

PR #11, opened 2019-11-11
Allowing for older versions of Faraday

cassiascheffer:
hm-proj requires an older version of Faraday. Allowing for version 0.9.0 as we don't have any features here that require a higher version.


harrylewis (approved):
We almost locked ourselves out of our own gem 😨

PR #12, opened 2019-11-14
Changing Class names so bundler auto-requires gem

cassiascheffer:
This PR modifies the structure of the gem a bit so that we don't have to add `require: bambora` when we include it in our Gemfile.

It also adds some utility methods to `Bambora::Client` so that it handles initializing resource classes.

*NOTE*: The change from `Bambora::Client` to `Bambora::RestClient` is a breaking change. I will update hm-proj in a later PR.


lib/bambora/client.rb
```diff
@@ -1,47 +1,101 @@
...
-require 'base64'
-require 'faraday'
+require 'bambora/rest_client'
+require 'bambora/headers'
+require 'bambora/json_client'
+require 'bambora/v1/batch_payment_report_resource'
+require 'bambora/v1/batch_payment_resource'
+require 'bambora/v1/profile_resource'
+require 'bambora/v1/payment_resource'
+require 'bambora/client/version'
```
harrylewis: Is a certain order required for loading/scoping purposes? Could this be alphabetized as before?

lib/bambora/client.rb
```diff
@@ -1,47 +1,101 @@
...
       yield(self) if block_given?
     end
 
-    protected
-
-    def get(path:, params:, headers:)
-      connection.get(path, params, headers)
+    # Retrieve a client to make requests the Profiles endpoints.
+    #
+    # @example
```
harrylewis: Small spacing/grammar quibble. Change "make requests the" to "make requests to the", and new line after `@example`?

lib/bambora/client.rb
```diff
@@ -1,47 +1,101 @@
...
-      end
-    end
+    def batch_payments; end
+
+    private
 
-    def build_headers(api_key)
-      Bambora::Headers.build(api_key: api_key, merchant_id: merchant_id, sub_merchant_id: sub_merchant_id)
+    def json_client
+      @json_client ||= Bambora::JSONClient.new(
```
harrylewis, line 91: I don't think any of this would have been possible without moving the API keys to the `*Resource` classes, so that was a great change!

lib/bambora/rest_client.rb
```diff
@@ -0,0 +1,55 @@
...
+    # Initialze a client that makes REST requests. The RestClient is used by the JSONClient.
+    #
+    # @example
+    #   rest_client = Bambora::RestClient.new(
+    #     base_url: ENV.fetch('BAMBORA_BASE_URL'),
+    #     merchant_id: ENV.fetch('BAMBORA_MERCHANT_ID'),
+    #     sub_merchant_id: ENV.fetch('BAMBORA_SUB_MERCHANT_ID'),
+    #   )
+    def initialize(options = {})
+      if !options[:version].nil? && options[:version].upcase != 'V1'
```
harrylewis: Not necessarily blocking, but this piece of code is now in a few places. Is it possible to turn this into a shareable piece of code across the gem?

lib/bambora/client.rb
```diff
@@ -1,47 +1,101 @@
...
+require 'bambora/rest_client'
+require 'bambora/headers'
+require 'bambora/json_client'
+require 'bambora/v1/batch_payment_report_resource'
+require 'bambora/v1/batch_payment_resource'
+require 'bambora/v1/profile_resource'
+require 'bambora/v1/payment_resource'
+require 'bambora/client/version'
 
 module Bambora
```
harrylewis, line 12: `Bambora::Client` is more of a convenience class, but would it be worth adding tests for it?

README.md
```diff
@@ -31,39 +31,34 @@ You can initilize the client by passing a hash to it, or by passing a block.
...
 
-The client is then passed to one of the `*Resource` classes, mentioned below, in order to make a network request.
-
 ### Profiles
 
 *Summary*: Payment profiles store confidential payment information. Transactions can be processed against profiles.
 *Bambora Docs*: <https://dev.na.bambora.com/docs/guides/payment_profiles/>
 
-To use the profiles API, you must create an instance of the `Bambora::V1::ProfileResource` class.
+To use the profiles API, you must create an instance of the `Bambora::Client::V1::ProfileResource` class.
```
harrylewis: Since this is the front of the gem that consumers will initially see, it might be more welcoming to strictly show the convenience class method, rather than the lower level methods. What do you think?

lib/bambora/client.rb
```diff
@@ -1,47 +1,101 @@
...
-      connection.get(path, params, headers)
+    # Retrieve a client to make requests the Profiles endpoints.
+    #
+    # @example
+    #   profiles = client.profiles
+    #   profiles.delete(customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A')
+    #
+    # @return [Bambora::V1::ProfileResource]
+    def profiles
+      @profiles ||= Bambora::V1::ProfileResource.new(client: json_client, api_key: profiles_api_key)
```
harrylewis: Is `::Client` missing here? And below?

harrylewis (commented):
Another great quality of life improvement, making consumption of this gem much easier. I had a few comments, will approve once they are addressed.

lib/bambora/client.rb
cassiascheffer: What do you mean here? `Change "make requests the" to "make requests to the"`

cassiascheffer: Oh gosh 🤦‍♀ Got it.

lib/bambora/client.rb
cassiascheffer: I ended up choosing to not name-space everything inside of `Client`.

lib/bambora/client.rb
cassiascheffer: Yes! This totally slipped my mind.

lib/bambora/client.rb
harrylewis: Sorry, that was a bit ambiguous because I commented over a set of lines.

lib/bambora/client.rb
```diff
@@ -1,47 +1,107 @@
...
+    #
+    # @example
+    #   profiles = client.profiles
+    #   profiles.delete(customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A')
+    #
+    # @param api_key [String] optional API key for the profiles endpoint. If the Client class was initialized with an
+    #   API key, this parameter is not needed.
+    #
+    # @return [Bambora::V1::ProfileResource]
+    def profiles(api_key: profiles_api_key)
```
harrylewis: Since the resource-specific API keys are now passed in for each resource method, are they still applicable at the class instantiation level?

README.md
```diff
@@ -108,15 +106,11 @@ profiles.delete(customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A')
 Android Pay.
 *Bambora Docs*: <https://dev.na.bambora.com/docs/references/payment_APIs/>
 
-To use the payments API, you must create an instance of the `Bambora::V1::PaymentResource` class.
-
 ```ruby
-client = Bambora::JSONClient.new(...)
-payments = Bambora::V1::PaymentResource.new(client: client, api_key: ENV.fetch('BAMBORA_PAYMENTS_API_KEY'))
+client = Bambora::Client.new(...)
+payments = client.payments
```
harrylewis: Is this able to be instantiated without an `api_key` parameter?

lib/bambora/client.rb
```diff
@@ -1,47 +1,107 @@
...
+    #       },
+    #     },
+    #   )
+    #
+    # @param api_key [String] optional API key for the payments endpoint. If the Client class was initialized with an
+    #   API key, this parameter is not needed.
+    #
+    # @return [Bambora::V1::PaymentResource]
+    def payments(api_key: payments_api_key)
+      @payments ||= Bambora::V1::PaymentResource.new(client: json_client, api_key: api_key)
```
harrylewis, line 81: Ah, I was confused about this, I thought you namespaced all the files under `/client`. I see now, that this class still lives in `/bambora/v1/...`.

lib/bambora/client.rb
```diff
@@ -1,47 +1,107 @@
...
+    # @param options[merchant_id] [String] The Merchant ID for this request.
+    # @param options[sub_merchant_id] [String] The Sub-Merchant ID for this request.
+    # @param options[profiles_api_key] [String] The Profiles API Key for this request.
+    # @param options[payments_api_key] [String] The Payments API Key for this request.
+    # @param options[batch_api_key] [String] The Batch API Key for this request.
+    # @param options[reporting_api_key] [String] The Reporting API Key for this request.
     def initialize(options = {})
-      unless options[:version].nil?
-        raise Bambora::Error, 'Only V1 endpoints are supported at this time.' if options[:version].upcase != 'V1'
+      if !options[:version].nil? && options[:version].upcase != 'V1'
```
sweetyclem, line 33: This condition is a little confusing. Maybe it could be easier to read with like ` if options[:version].present?`

harrylewis (commented):
I really like the addition of the `:api_key` parameter for the `Bambora::Client` methods. It seems to scope everything really nicely. With that in mind, are the class-level equivalent parameters necessary?

sweetyclem (approved):

lib/bambora/client.rb
cassiascheffer: I figured that could be up to the user. They can either initilize with the API keys they need, or pass the key in when they ask for the class.

lib/bambora/client.rb
cassiascheffer: `.present?` only exists in Rails! Writing Ruby without Rails makes you appreciate all the nice things Rails does for you.

harrylewis (approved):
Looks great! Thanks for all this work, the gem will be so much easier to consume now.

PR #13, opened 2019-11-21
Adding Clients and Adapters for Legacy APIs

cassiascheffer:
This is all setup for interacting with the legacy APIs located under the `/scripts` path.

- Adds XML Client for making requests with an XML body
- Adds JSONResponse for parsing JSON response bodies
- Adds QueryStingResponse for parsing query string response bodies
- Adds Client for making POST requests with a query string


lib/bambora/client.rb
```diff
@@ -1,20 +1,28 @@
...
 
 module Bambora
+  ##
+  # The Client class is used to initialize Resource objects that can make requests to the Bambora API.
   class Client
     class Error < StandardError; end
-    class ServerError < StandardError; end
 
-    attr_accessor :base_url, :merchant_id, :sub_merchant_id
+    attr_reader :base_url, :merchant_id, :sub_merchant_id
```
cassiascheffer, line 42: These should have been read-only all along.

lib/bambora/headers.rb
```diff
@@ -1,17 +1,30 @@
...
+  ##
+  # Builds Headers for HTTP requests.
   class Headers
-    class << self
-      def build(api_key:, merchant_id:, sub_merchant_id: nil)
-        headers = { 'Authorization' => "Passcode #{passcode(merchant_id, api_key)}" }
-        headers['Sub-Merchant-Id'] = sub_merchant_id unless sub_merchant_id.nil?
-        headers
-      end
+    attr_reader :api_key, :merchant_id, :sub_merchant_id, :content_type
```
cassiascheffer, line 7: The class method here started to have too many arguments, so I turned it into an instance instead.

lib/bambora/json_client.rb
```diff
@@ -26,9 +30,9 @@ class JSONClient < Bambora::RestClient
     #
     # @return [Hash] Indicating success or failure of the operation.
     def get(path:, params: nil, api_key:)
-      parse_response(
-        super(path: path, params: params, headers: build_headers(api_key)),
-      )
+      Bambora::JSONResponse.new(
```
cassiascheffer: Needed to share some methods, so now the response is parsed by its own class.

lib/bambora/xml_client.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    # Make a POST Request.
+    #
+    # @param path [String] Indicating request path.
+    # @param body [Hash] Data to be sent in the body of the request.
+    # @param api_key [String] Indicating the API Key to be used with the request.
+    #
+    # @return [Hash] Indicating success or failure of the operation.
+    def post(path:, body:, api_key:)
+      Bambora::ResponseAdapterFactory.for(
```
cassiascheffer: We can get JSON, query strings, or XML back from these endpoints. query strings are easier to parse in Ruby. The endpoints that give us an option have either XML/JSON, or XML/query string. I don't think we'll need to add an XML parser, but we can add one if we need to.

lib/bambora/www_form_client.rb
```diff
@@ -0,0 +1,32 @@
...
+    #
+    # @param path [String] Indicating request path.
+    # @param body [Hash] Data to be sent in the query parameters of the request.
+    # @param api_key [String] Indicating the API Key to be used with the request.
+    #
+    # @return [Hash] Indicating success or failure of the operation.
+    def post(path:, body:, api_key:)
+      Bambora::ResponseAdapterFactory.for(
+        super(
+          path: "#{path}?#{WWWFormParameters.new(body: body)}",
```
cassiascheffer, line 19: Excon's and Faraday's docs show that you can pass query parameters into the `body` and they get formatted into the URL, but this wasn't working for me, so I'm formatting them here and tacking them onto the end of the URL.

spec/bambora/www_form_client_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+            .to_return(headers: { 'Content-Type' => 'text/html' }, body: query_string_response_body)
+        end
+
+        it 'parses the response' do
+          resp = subject.post(path: path, body: request_body, api_key: api_key)
+          expect(resp).to eq response_body
+        end
+      end
+
+      # context 'server responds with a non 2xx status' do
```
sweetyclem: Why is this code commented out? Is it something that could be useful and we don't want to delete?

sweetyclem (approved):
Looks good from the little I can understand 😄

spec/bambora/www_form_client_spec.rb
cassiascheffer: Ah, oops! I had commented this out while debugging. Fixed!

lib/bambora/client.rb
```diff
@@ -30,10 +55,6 @@ class ServerError < StandardError; end
     # @param options[merchant_id] [String] The Merchant ID for this request.
     # @param options[sub_merchant_id] [String] The Sub-Merchant ID for this request.
     def initialize(options = {})
-      if !options[:version].nil? && options[:version].upcase != 'V1'
```
harrylewis, line 33: Was this check moved somewhere else?

lib/bambora/client.rb
```diff
@@ -1,20 +1,45 @@
...
+require 'json'
+require 'faraday'
+require 'gyoku'
+
 require 'bambora/client/version'
+
+# Builders
 require 'bambora/headers'
+require 'bambora/xml_request_body'
+require 'bambora/www_form_parameters'
```
harrylewis, line 14: Love how you've organized these dependencies. If gives a really clear structure to all the classes, and help future contributors to easily see where their contribution might fit in.

lib/bambora/headers.rb
```diff
@@ -1,17 +1,30 @@
...
 
-      def passcode(merchant_id, api_key)
-        Base64.encode64("#{merchant_id}:#{api_key}").delete("\n")
+    def initialize(options = {})
+      options.each do |key, value|
+        instance_variable_set("@#{key}", value)
       end
     end
+
+    def build
```
harrylewis, line 26: More of a general question, when dealing with the Builder Pattern, is it usually expected to have a method called something like `#build`? That way, with any builder you can instantiate the instance, and "universally" call `#build` to get the intended result?

lib/bambora/headers.rb
harrylewis: At first, I was confused by this change, because it just seems to shift the long list of parameters from the method to the class instantiation, but I see it is encapsulated by the **single** `:options` parameter, which follows [introduce parameter object](https://refactoring.guru/introduce-parameter-object) 🙌.

lib/bambora/headers.rb
```diff
@@ -1,17 +1,30 @@
...
-      def build(api_key:, merchant_id:, sub_merchant_id: nil)
-        headers = { 'Authorization' => "Passcode #{passcode(merchant_id, api_key)}" }
-        headers['Sub-Merchant-Id'] = sub_merchant_id unless sub_merchant_id.nil?
-        headers
-      end
+    attr_reader :api_key, :merchant_id, :sub_merchant_id, :content_type
 
-      def passcode(merchant_id, api_key)
-        Base64.encode64("#{merchant_id}:#{api_key}").delete("\n")
+    def initialize(options = {})
```
harrylewis, line 16: Does this need method documentation to describe the allowed parameters?

lib/bambora/json_client.rb
```diff
@@ -66,9 +70,9 @@ def get(path:, params: nil, api_key:)
     #
     # @return [Hash] Indicating success or failure of the operation.
     def post(path:, body:, api_key:)
-      parse_response(
-        super(path: path, body: body.to_json.to_s, headers: build_headers(api_key)),
-      )
+      Bambora::JSONResponse.new(
```
harrylewis: I've used method decorators quite a bit in the past with Python. I'm not sure if they are used much in Ruby, but since all of the method returns here are encapsulated by this class instantiation and call, it might be an interesting thing to add to DRY the code up, as long as it maintains readability. Nothing to do necessarily here.

lib/bambora/v1/profile_resource.rb
```diff
@@ -14,13 +17,14 @@ class ProfileResource
...
       #
-      # @param client [Bambora::RestClient] An instance of Bambora::JSONClient, used to make network requests.
+      # @param client [Bambora::JSONClient] An instance of Bambora::JSONClient, used to make network requests.
       def initialize(client:, api_key:)
         @client = client
         @api_key = api_key
         @sub_path = '/v1/profiles'
       end
 
+      ##
```
harrylewis, line 27: Is this `rdoc`/`yard` convention to prefix the docstring with a `##` line?

lib/bambora/www_form_client.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Bambora
+  # The base class for making XML requests.
```
harrylewis: Is `XML` here a typo?

lib/bambora/www_form_parameters.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Bambora
+  ##
+  # Builds WWW URL Encoded request parameters from a Hash
+  class WWWFormParameters
+    attr_reader :body
+
+    def initialize(body:)
```
harrylewis, line 13: What should `:body` look like? Maybe add method documentation so this is clear?

lib/bambora/www_form_parameters.rb
```diff
@@ -0,0 +1,17 @@
...
+      @body = body
+    end
+
+    def to_s
```
harrylewis, line 21: The `#to_s` naming feels a little ambiguous to me. What about something like `#to_encoded_s`?

harrylewis: Looking at the other request builder methods, it seems like it would follow the adapter pattern more closely to align all of the method namings, as you have done, so no issue here.

lib/bambora/adapters/query_string_response.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Bambora
+  ##
+  # Parses a JSON response into a Hash
```
harrylewis: Typo here on `JSON`?

lib/bambora/adapters/query_string_response.rb
```diff
@@ -0,0 +1,20 @@
...
+  class QueryStringResponse < Response
+    def to_h
+      parsed_response = super
```
harrylewis, line 8: Will a returned query response be an Array?

lib/bambora/json_client.rb
harrylewis: Could `ResponseAdapterFactory` be used here?

spec/bambora/adapters/json_response_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    describe '#to_h' do
+      context 'a valid response' do
+        let(:json_response_body) { { gelflings: %w[rian deet brea] } }
+        let(:faraday_response) { Faraday::Response.new(body: json_response_body.to_json.to_s) }
+        subject { described_class.new(faraday_response).to_h }
+        it { is_expected.to eq json_response_body }
+      end
+
+      context 'an invalid response' do
+        let(:string_response_body) { 'GARTHIM! ATTACK!' }
```
harrylewis, line 16: I don't know the reference but it's still funny 😆.

spec/bambora/xml_client_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    let(:response_body) { { response: 'body', with: { objects: 'yay!' }, and: [{ arrays: 'wow!' }] } }
+    let(:failed_status) { 500 }
+    let(:failed_response_body) { 'Mouldy mildew, mother of mouthmuck, dangle and strangle and death.' }
+
+    subject { described_class.new(base_url: base_url, merchant_id: merchant_id) }
+
+    describe '#post' do
+      let(:request_body) { { gelfling: 'Deet' } }
+      let(:xml_request_body) do
+        "<?xml version='1.0' encoding='utf-8'?>#{Gyoku.xml(request: request_body.merge(rpt_format: 'JSON'))}"
```
harrylewis, line 26: Does using `Gyoku` directly again here create a sort of self-serving test? It seems to mirror the implementation pretty closely. Maybe it would be best to explicitly unpack this XML structure?

spec/support/shared_examples/json_response_parser_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+shared_examples 'a json response parser' do
```
harrylewis, line 3: Using all of the things in the gem, including `shared_examples`. I love it!

harrylewis (commented):
That was a doozy! Very cool to see the evolution of it, and nice to see all the use of design patterns here. Lots of learning for me, that's for sure.

I had some comments throughout, not all need to be acted upon though.

lib/bambora/client.rb
cassiascheffer: We don't actually have it documented and there's no `attr_accessor` for it, so I removed it. This change probably should have been in a separate PR.

lib/bambora/headers.rb
cassiascheffer: Yes, it is. Although, I normally see it as a class method. I tried both and the instance method gave me a tidier class. Passing a parameter object into a `.build` class resulted in quite a long method. This might be unconventional for a builder though.

lib/bambora/v1/profile_resource.rb
cassiascheffer: I think a linter I installed started telling me I should do this. I think I got carried away though.

lib/bambora/adapters/query_string_response.rb
cassiascheffer: It's a hash. I'm not a fan of this adapter. It kind of fudges what the response could be and assumes that having an empty array for the values is a parsing error.

Basically, if you do `CGI.parse('somestring')` you get `{ 'somestring' => [] }` back. The check on the next line is to see if there are empty values here.

Now that I think about this, though, would it be possible for the API to rerun a query string like so `someParam=&someOtherParam=value`? In this case, we'd unintentionally raise an error.

Let me do some research on this one and come back to it.

lib/bambora/json_client.rb
cassiascheffer: It could, but the response will always be JSON here.

lib/bambora/json_client.rb
cassiascheffer: Can you explain to me what this would look like in Ruby? Perhaps we can chat offline.

spec/bambora/adapters/json_response_spec.rb
cassiascheffer: The best 80s movie, which is now a Netflix series! https://www.youtube.com/watch?v=AEH36q8PiXo

spec/bambora/xml_client_spec.rb
cassiascheffer: Yeah, I was thinking about that when I wrote this. My assumption was that Gyoku itself _should_ have tests to ensure it behaves correctly.

lib/bambora/adapters/json_response.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Bambora
+  ##
+  # Parses a JSON response into a Hash
+  class JSONResponse < Response; end
```
erawk, line 6: I know you are explicitly `require`ing this and other adapters. Can you tell me some of the reasoning behind not adding in the `Bambora::Adapters` class namespace? I like that it's fewer chars to type 😄

lib/bambora/www_form_client.rb
erawk: Is it worth adding this comment in the code as to _why_ you're building the query parameter string?

erawk (approved):
Two questions around scoping classes and documentation. This looks great, so I'd 🚢 this version unless you feel strongly about either of the comments!

lib/bambora/json_client.rb
```diff
@@ -89,38 +93,15 @@ def post(path:, body:, api_key:)
     #
     # @return [Hash] Indicating success or failure of the operation.
     def delete(path:, api_key:)
-      parse_response(
-        super(path: path, headers: build_headers(api_key)),
-      )
+      parse_response_body(
+        super(path: path, headers: build_headers(api_key: api_key)),
+      ).to_h
```
harrylewis, line 98: I think the `#to_h` is unnecessary here, but I don't think `{}.to_h.to_h` will cause an exception or return incorrect data.

harrylewis (approved):
This looks great, let's ship it and start using it!

lib/bambora/adapters/json_response.rb
cassiascheffer: Mostly because I had written this class before I moved all this into a directory. I considered adding a namespace, but I was focused on getting this feature done. Happy to add this as a tidy up later.

lib/bambora/json_client.rb
cassiascheffer: Right right. I'm going to merge this and add a follow up to get rid of the extra `to_h`.

lib/bambora/www_form_client.rb
cassiascheffer: Yes! I'd like to get this PR merged though, so I'll add that in a follow up.

PR #14, opened 2019-11-29
Feat/bank profile api

cassiascheffer:
This PR adds the ability to create bank profiles through the `/scripts/payment_profiles.asp` endpoint, as documented here: https://help.na.bambora.com/hc/en-us/articles/115010346067-Secure-Payment-Profiles-Batch-Payments


lib/bambora/bank/adapters/payment_profile_response.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Bambora
+  module Bank
```
cassiascheffer, line 4: This adapter is very specific to the bank profile endpoint, so I wanted to keep it in the bank name space.

lib/bambora/client.rb
```diff
@@ -20,6 +22,7 @@
 require 'bambora/adapters/response'
 require 'bambora/adapters/json_response'
 require 'bambora/adapters/query_string_response'
+require 'bambora/bank/adapters/payment_profile_response'
```
cassiascheffer, line 27: You'll notice that other "builders" and "adapters" don't have nice name spaces like these ones do. I'm going to do a follow up PR that re organizes this stuff a bit.

lib/bambora/factories/response_adapter_factory.rb
```diff
@@ -8,8 +8,11 @@ class << self
       def for(response)
         content_type = response.headers['Content-Type']
         case content_type
-          when 'application/json' then Bambora::JSONResponse.new(response)
-          when 'text/html' then Bambora::QueryStringResponse.new(response)
+          when 'application/json'
+            Bambora::JSONResponse.new(response)
+          when 'text/html'
+            # Currently, the only endpoint that responds wit text/html is /scripts/payment_profiles.asp
```
cassiascheffer, line 14: As we build other endpoints, we might need to change the case statement here to look at request URLs instead of content type. But for now, this works fine.

lib/bambora/www_form_client.rb
```diff
@@ -13,20 +13,25 @@ class WWWFormClient < Bambora::RestClient
     # @param api_key [String] Indicating the API Key to be used with the request.
     #
     # @return [Hash] Indicating success or failure of the operation.
-    def post(path:, body:, api_key:)
+    def post(path:, body:)
```
cassiascheffer, line 15: Couple of small changes here that I didn't account for in my first PR.

lib/bambora/www_form_client.rb
```diff
@@ -13,20 +13,25 @@ class WWWFormClient < Bambora::RestClient
     # @param api_key [String] Indicating the API Key to be used with the request.
```
sweetyclem: This seems to have been removed, should probably remove the comment too.

lib/bambora/www_form_client.rb
```diff
@@ -13,20 +13,25 @@ class WWWFormClient < Bambora::RestClient
...
+    def post(path:, body:)
+      # Both Faraday's and Excon's docs show that you can pass a hash into the +body+ and set the content type to
+      # application/x-www-form-urlencoded and the +body+ will be transformed into query parameters, however, this
+      # did not work in testing so I am manually transforming the hash into query parameters here.
       parse_response_body(
         super(
           path: "#{path}?#{WWWFormParameters.new(body: body)}",
           body: nil,
-          headers: build_headers(api_key: api_key),
+          headers: build_headers,
```
sweetyclem, line 23: Why aren't we using the `api_key` anymore?

spec/bambora/bank/adapters/payment_profile_response_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+        describe '#to_h' do
+          context 'a valid response' do
+            let(:json_response_body) { { someGelflings1: %w[rian deet brea] } }
+            let(:query_string_response_body) { URI.encode_www_form(json_response_body) }
+            let(:faraday_response) { Faraday::Response.new(body: query_string_response_body) }
+            subject { described_class.new(faraday_response).to_h }
+            it { is_expected.to eq some_gelflings_1: %w[rian deet brea] }
+          end
+
+          context 'an invalid response' do
```
sweetyclem: The new rubocop linting rules are always telling me to add `with` or `when` to a context description, could you check?

spec/bambora/bank/payment_profile_resource_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+        end
+
+        before do
+          stub_request(:post, "#{base_url}/scripts/payment_profiles.asp").with(
+            query: query_params,
+            headers: headers,
+          ).to_return(headers: response_headers, body: response_body)
+        end
+
+        it "POST's to the Bambora API" do
```
sweetyclem, line 109: We could use single quotes here for convention

sweetyclem (commented):
Looks great! I only have minor suggestions

spec/bambora/bank/adapters/payment_profile_response_spec.rb
harrylewis: @sweetyclem with the current `rubocop` version that is being used in this repository, we would have to manually add [`rubocop-rspec`](https://github.com/rubocop-hq/rubocop-rspec) to the dependencies to do spec linting.

This is something that is definitely worthwhile, to allow consistency on the spec writing. I like the guideline you mentioned!

spec/bambora/bank/payment_profile_resource_spec.rb
harrylewis: @sweetyclem I believe the double quotes here was used because the text contains a single quote inside it. I think with `rubocop`, is there is string interpolation involved, or a single quote within the string, then double quotes are accepted.

lib/bambora/bank/payment_profile_resource.rb
```diff
@@ -0,0 +1,74 @@
...
+      DEFAULT_VERSION = 1.0
+      attr_reader :client, :api_key, :sub_path, :version
+
+      ##
+      # Instantiate an interface to make requests against Bambora's Profiles API.
+      #
+      # @example
+      #
+      #   client = Bambora::WWWFormClient(base_url: '...', api_key: '...', merchant_id: '...')
+      #   profiles = Bambora::Bank::PaymentProfileResource(client: client)
```
harrylewis: In the documentation, should the resource be passed the `api_key`, instead of the client?

spec/bambora/bank/payment_profile_resource_spec.rb
sweetyclem: Ohh I didn't realize this was not in `hm-proj`!

spec/bambora/bank/adapters/payment_profile_response_spec.rb
sweetyclem: Ohh I didn't realize this was not in hm-proj!

lib/bambora/bank/payment_profile_resource.rb
```diff
@@ -0,0 +1,74 @@
...
+      #   # Start making requests ...
+      #
+      # @param client [Bambora::JSONClient] An instance of Bambora::JSONClient, used to make network requests.
+      # @param api_key [String] An API key for this endpoint. This is also known as the "Pass Code"
+      # @param version [String] The Service Version you are requesting from the server.
+      def initialize(client:, api_key:)
+        @client = client
+        @api_key = api_key
+        @sub_path = '/scripts/payment_profiles.asp'
+        @version = version || DEFAULT_VERSION
```
harrylewis: If `version` isn't defined anywhere, can this just be `@version = DEFAULT_VERSION`? Actually, it's in the initializer documentation, so maybe it should be in the parameter list?

lib/bambora/bank/payment_profile_resource.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+
+      ##
+      # Create a Bank Payment Profile
+      #
+      # @example
+      #  data = {
+      #   customer_code: '1234',
```
harrylewis: The indentation in this example seems a bit funny, could it be expanded to be 2 spaces?

lib/bambora/client.rb
```diff
@@ -115,5 +151,13 @@ def json_client
         sub_merchant_id: sub_merchant_id,
       )
     end
+
+    def www_form_client
+      @www_form_client ||= Bambora::WWWFormClient.new(
+        base_url: base_url,
```
harrylewis, line 157: Not a blocker, it looks like the arguments passed in for the `#json_client` and the `#www_form_client` are the same. Maybe they could be moved to method to DRY it up?

lib/bambora/www_form_client.rb
cassiascheffer: The API key gets merged into the `body` for this request before we pass them into the client here.

lib/bambora/client.rb
```diff
@@ -94,14 +98,46 @@ def profiles(api_key:)
     #     },
     #   )
     #
-    # @param api_key [String] optional API key for the payments endpoint. If the Client class was initialized with an
-    #   API key, this parameter is not needed.
+    # @param api_key [String] optional API key for the payments endpoint.
```
harrylewis: Thanks for removing this outdated comment. Remove it on line 75-76 as well?

spec/bambora/bank/adapters/payment_profile_response_spec.rb
cassiascheffer: Oh, I like `when`. I'll add that.

lib/bambora/client.rb
```diff
@@ -94,14 +98,46 @@ def profiles(api_key:)
...
+    #   profiles = client.bank_profiles
+    #
+    #   data = {
+    #     customer_code: '1234',
+    #     bank_account_type: 'CA',
+    #     account_holder: account_holder,
+    #     institution_number: '123',
+    #     branch_number: '12345',
+    #     account_number: '123456789',
+    #     name: name,
```
harrylewis: I assume this was pulled straight from an example you had wrote in development. Is it overkill to include an explicit example of each of the fields, rather than a variable name?

spec/bambora/bank/payment_profile_resource_spec.rb
cassiascheffer: Yeah, this is my least favourite part of the single quote rule. It'd be better to use double quotes everywhere, I think, but them's the rules.

lib/bambora/client.rb
```diff
@@ -94,14 +98,46 @@ def profiles(api_key:)
...
+    #     postal_code: postal_code,
+    #     province: province,
+    #     country: country,
+    #     sub_merchant_id: sub_merchant_id,
+    #     operation_type: 'N',
+    #   }
+    #
+    #   profiles.create(data)
+    #
+    # @param api_key [String] optional API key for the bank profiles endpoint.
```
harrylewis: Are these `api_key` parameters still optional without the class level initializer?

lib/bambora/factories/response_adapter_factory.rb
harrylewis: I'm all for this iterative approach.

lib/bambora/www_form_client.rb
harrylewis: This endpoint is so bizarre. I guess that's legacy code for you 🤷‍♂.

lib/bambora/client.rb
harrylewis: Again, I love that all these patterns are being used here!

lib/bambora/bank/payment_profile_resource.rb
```diff
@@ -0,0 +1,74 @@
...
+      #   account_number: '123456789',
+      #   name: 'Hup Podling',
+      #   email_address: 'Brea Princess of Vapra',
+      #   phone_number: '1231231234',
+      #   address_1: 'The Castle',
+      #   city: "Ha'rar",
+      #   postal_code: 'H0H 0H0',
+      #   province: 'Vapra',
+      #   country: 'Thra',
+      #   sub_merchant_id: '1',
```
harrylewis: Is `sub_merchant_id` necessary? It looks like it will always be overridden by `#payment_profile_body`.

lib/bambora/bank/payment_profile_resource.rb
```diff
@@ -0,0 +1,74 @@
...
+      #  }
+      #
+      #  payment_profile_resource.create(data)
+      #
+      #  @params profile_data [Hash] with values as noted in the example.
+      def create(profile_data)
+        client.post(path: sub_path, body: payment_profile_body(profile_data))
```
harrylewis, line 61: The example above is missing an `operationType`, which in the case of creation should be `N`. Could this be hard-coded into the implementation?

lib/bambora/bank/payment_profile_resource.rb
```diff
@@ -0,0 +1,78 @@
...
+
+module Bambora
+  module Bank
+    ##
+    # For making requests to the /scripts/payment_profiles.asp endpoint
+    #
+    # @see https://help.na.bambora.com/hc/en-us/articles/115010346067-Secure-Payment-Profiles-Batch-Payments
+    class PaymentProfileResource
+      DEFAULT_VERSION = 1.0
+      DEFAULT_RESPONSE_FORMAT = 'QS'
```
harrylewis, line 11: Thanks for adding this!

lib/bambora/bank/builders/payment_profile_params.rb
```diff
@@ -0,0 +1,38 @@
...
+          # @params params [Hash]
+          def build(params)
+            params.each_with_object({}) do |(key, value), obj|
+              obj[transform_key(key)] = value
+            end
+          end
+
+          private
+
+          def transform_key(key)
```
harrylewis, line 25: Back to the basics!

spec/bambora/bank/builders/payment_profile_params_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+              'merchantId' => 1,
+              'subMerchantId' => 1,
+              'serviceVersion' => '1.0',
+              'operationType' => 'N',
+            }
+          end
+
+          subject { described_class.build(original_hash) }
+
+          it { is_expected.to eq expected_hash }
```
harrylewis, line 58: What a satisfying test.

spec/bambora/bank/builders/payment_profile_params_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+              'accountHolder' => 'All-Maudra Mayrin',
+              'institutionNumber' => '123',
+              'branchNumber' => '12345',
+              'accountNumber' => '123456789',
+              'ordName' => 'Hup Podling',
+              'ordEmailAddress' => 'Brea Princess of Vapra',
+              'ordPhoneNumber' => '1231231234',
+              'ordAddress1' => 'The Castle',
+              'ordCity' => "Ha'rar",
+              'ordPostalCode' => 'H0H 0H0',
```
harrylewis, line 46: 🎶 It's beginning to look a lot like Christmas ...

lib/bambora/bank/payment_profile_resource.rb
```diff
@@ -0,0 +1,78 @@
...
+      # Instantiate an interface to make requests against Bambora's Profiles API.
+      #
+      # @example
+      #
+      #   client = Bambora::WWWFormClient(base_url: '...', api_key: '...', merchant_id: '...')
+      #   profiles = Bambora::Bank::PaymentProfileResource(client: client)
+      #
+      #   # Start making requests ...
+      #
+      # @param client [Bambora::JSONClient] An instance of Bambora::JSONClient, used to make network requests.
```
harrylewis: I think this should be `Bambora::WWWFormClient`.

lib/bambora/bank/adapters/payment_profile_response.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        private
+
+        def parsed_query_string
+          Bambora::QueryStringResponse.new(response).to_h
+        end
+
+        def transform(camel_case_word)
+          word = camel_case_word.to_s
+          word.gsub!(/([a-z])([A-Z\d])/, '\1_\2')
```
harrylewis, line 29: That's some regex 👀. Cool usage of regex groupings.

harrylewis (commented):
The case converters were really cool to look at, and you've really brought a lot of elegance to a clumsy endpoint. I had a few comments sprinkled, nothing that warrants any major refactoring though.

lib/bambora/bank/payment_profile_resource.rb
```diff
@@ -0,0 +1,78 @@
...
+      # @param api_key [String] An API key for this endpoint. This is also known as the "Pass Code"
+      # @param version [String] The Service Version you are requesting from the server.
+      def initialize(client:, api_key:)
```
erawk: Is `version` missing as a keyword argument here?

lib/bambora/bank/builders/payment_profile_params.rb
```diff
@@ -0,0 +1,38 @@
...
+  module Bank
+    module Builders
+    ##
+    # Builds a request body for the Bank Payment Profile endpoint from a Hash
+      class PaymentProfileParams
+        CONTACT_PARAMS = %w[name email_address phone_number address_1 address_2 city postal_code province country].freeze
+
+        class << self
+          ##
+          # Converts a snake_case hash to camelCase keys with vendor-specific previxes.
```
erawk: Small typo

```suggestion
          # Converts a snake_case hash to camelCase keys with vendor-specific prefixes.
```

lib/bambora/client.rb
```diff
@@ -3,6 +3,7 @@
 # Libraries
 require 'base64'
 require 'json'
+require 'cgi'
```
erawk, line 5: Did you want to alphabetize this list? Does it make sense to separate Ruby standard library `require`'s from others?

erawk (commented):
Nice work! Couple minor comments.

Not sure if this is on the roadmap, but is the idea to generate static docs using many of the helpful `rubydoc` prefix content in the codebase?

lib/bambora/client.rb
cassiascheffer: Ah, meant to do this but was focused on functionality. I'll update it!

lib/bambora/client.rb
cassiascheffer: They're not! I'll change this.

lib/bambora/bank/payment_profile_resource.rb
cassiascheffer: Ah, yes yes. This should be hard-coded into the implementation!

lib/bambora/bank/adapters/payment_profile_response.rb
cassiascheffer: Not gonna lie, I looked at the Rails regex for this to be sure I didn't miss anything.

lib/bambora/bank/payment_profile_resource.rb
cassiascheffer: It is!

lib/bambora/client.rb
cassiascheffer: Yeah, I'll separate standard lib from others. I didn't alphabetize originally because load order matters for classes in this gem. But alphabetizing here makes sense.

cassiascheffer:
@erawk Yeah, Harry and I had talked about using yard for documentation. Most of the docs I've added are copy-pasted from the tests or from similar methods (hence me missing some things!).

lib/bambora/client.rb
cassiascheffer: I am going to leave this one as is because using a hash in place of key word arguments will be deprecated in 2.7. I don't want to block support for future versions of ruby.

https://bugs.ruby-lang.org/issues/14183

cassiascheffer: Ah, I just realized that the Clients use an options hash and not key word args, so my comment about Ruby 2.7 doesn't apply here. I'll consider moving this options hash to a method in a future PR.

lib/bambora/bank/payment_profile_resource.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      def payment_profile_body(profile_data)
+        Bambora::Bank::Builders::PaymentProfileParams.build(
+          profile_data.merge(
+            pass_code: api_key,
+            merchant_id: client.merchant_id,
+            sub_merchant_id: client.sub_merchant_id,
+            service_version: version,
+            response_format: DEFAULT_RESPONSE_FORMAT,
+            operation_type: CREATE_OPERATION_TYPE,
```
harrylewis, line 74: Not a blocker, but if we add more operations/endpoints in the future, we may have to put the merging of this key-value pair in the `#create` method. But since we only have one endpoint at the moment, I'm cool to leave this here!

spec/bambora/bank/payment_profile_resource_spec.rb
harrylewis: I *think* there is a slight performance improvement using single quotes, because you can guarantee there is no interpolation, and removes a few processing steps.

harrylewis (approved):
Can't wait to use this in our codebase! So much cleaner now, and most importantly, thread-safe.

lib/bambora/bank/payment_profile_resource.rb
cassiascheffer: That was my thinking here!

PR #16, opened 2019-11-29
Checking only content type. Excluding charset

cassiascheffer:
The JSON endpoints return the following content type: `application/json;
charset=utf-8`. We only care about the first part and not the charset.

THis raised an error when I tested the gem in staging.

### DIFF
--- a/lib/bambora/factories/response_adapter_factory.rb
+++ b/lib/bambora/factories/response_adapter_factory.rb
@@ -6,7 +6,7 @@ module Bambora
   class ResponseAdapterFactory
     class << self
       def for(response)
-        content_type = response.headers['Content-Type']
+        content_type = response.headers['Content-Type'].split(';').first
         case content_type
           when 'application/json' then Bambora::JSONResponse.new(response)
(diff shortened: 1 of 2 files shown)


lib/bambora/factories/response_adapter_factory.rb
```diff
@@ -6,7 +6,7 @@ module Bambora
   class ResponseAdapterFactory
     class << self
       def for(response)
-        content_type = response.headers['Content-Type']
+        content_type = response.headers['Content-Type'].split(';').first
```
harrylewis, line 9: I don't know much about content types, do they all follow the same format, delimited by colons, with the first argument being the overarching name?

harrylewis (approved):
Looks good to me!

lib/bambora/factories/response_adapter_factory.rb
cassiascheffer: Yeah, I forgot that they can include encoding data too. https://tools.ietf.org/html/rfc2616#section-7.2.1

PR #15, opened 2019-11-29
Restructuring files to use consistent name space

cassiascheffer:
This doesn't include any actual code changes. I've moved some files around so that class and file naming is consistent across the project.

Names like `Bambora::WWWFormParams` are now `WWW::Builders::WWWFormParams` so that it's more clear what the class does.

PR #17, opened 2019-12-03
Reorganize Builders and Clients into Name Spaces

cassiascheffer:
This PR accomplishes what #15 was supposed to do. I must have messed up when I did the rebase on that PR.

All "clients" are now in a "Rest" names space and all "builders" in a "Builders" namespace.


lib/bambora/builders/www_form_parameters.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Bambora
+  module Builders
+    ##
+    # Builds WWW URL Encoded request parameters from a Hash
+    class WWWFormParameters
```
erawk, line 7: Would this be `WwwFormParameters`? How does the autoloader work with multiple capitalized characters in a row?

lib/bambora/rest/client.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+module Bambora
+  module Rest
```
erawk, line 4: I don't have a preference, but am commenting on the inconsistency of using `REST` vs `Rest` compared with `WWW` -> `Www` and `XML` -> `Xml`. Looks like everything is explicitly required so I think it's all good, but maybe this should be REST? I assume this has to do with constants?

erawk (approved):
This looks great. Thanks for adding the extra namespacing!

lib/bambora/builders/www_form_parameters.rb
harrylewis: I believe as long as you `require` the correct file, the classes in that file will be visible. The autoloader should only apply to Rails?

lib/bambora/client.rb
```diff
@@ -11,26 +11,26 @@
...
 require 'bambora/adapters/response'
 require 'bambora/adapters/json_response'
 require 'bambora/adapters/query_string_response'
 require 'bambora/bank/adapters/payment_profile_response'
 
+# Builders
+require 'bambora/builders/headers'
+require 'bambora/builders/xml_request_body'
+require 'bambora/builders/www_form_parameters'
+require 'bambora/bank/builders/payment_profile_params'
```
harrylewis, line 24: Subtle difference here with the namespacing of the bank builders. Is the thinking that this is a separate domain, and could be moved around more easily if everything lives under `/bank`?

harrylewis (approved):
This looks great, and I'm guessing it deploys! 🎉 A few comments below, but no blockers.

lib/bambora/rest/client.rb
cassiascheffer: Ha, I thought about this. I chose `Rest` because it should technically be `ReST`, which I think looks weird.

lib/bambora/client.rb
cassiascheffer: Yeah, the bank stuff is a very different domain, so I chose to keep it all in that namespace. If I put it in the regular `builders` directory, the name `payment_profile_params` would easily be confused with the `v1/profiles` endpoint.

lib/bambora/rest/client.rb
cassiascheffer: Funny though, that I called it `REST` in one of my commits. I guess I don't have a strong opinion on this other than I wish the E stood for something.

PR #18, opened 2019-12-10
Using correct URI path, using correct params

cassiascheffer:
I had made two mistakes in my previous work. The URI should be singular,
and `account_holder` needs the `bank_` prefix.

### DIFF
--- a/lib/bambora/bank/payment_profile_resource.rb
+++ b/lib/bambora/bank/payment_profile_resource.rb
@@ -30,7 +30,7 @@ module Bambora
         @client = client
         @api_key = api_key
         @version = version
-        @sub_path = '/scripts/payment_profiles.asp'
+        @sub_path = '/scripts/payment_profile.asp'
       end
 
       ##
@@ -40,7 +40,7 @@ module Bambora
       #   data = {
       #     customer_code: '1234',
       #     bank_account_type: 'CA',
-      #     account_holder: 'All-Maudra Mayrin',
+      #     bank_account_holder: 'All-Maudra Mayrin',
       #     institution_number: '123',
       #     branch_number: '12345',
       #     account_number: '123456789',
--- a/lib/bambora/client.rb
+++ b/lib/bambora/client.rb
@@ -45,7 +45,7 @@ module Bambora
   class Client
     class Error < StandardError; end
 
-    attr_reader :base_url, :merchant_id, :sub_merchant_id
+    attr_reader :base_url, :merchant_id, :scripts_api_base_url, :sub_merchant_id
 
     # Initialze a new Bambora::Client.
     #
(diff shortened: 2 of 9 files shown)


lib/bambora/bank/payment_profile_resource.rb
```diff
@@ -30,7 +30,7 @@ def initialize(client:, api_key:, version: DEFAULT_VERSION)
         @client = client
         @api_key = api_key
         @version = version
-        @sub_path = '/scripts/payment_profiles.asp'
+        @sub_path = '/scripts/payment_profile.asp'
```
cassiascheffer, line 33: Changed this to singular.

lib/bambora/bank/payment_profile_resource.rb
```diff
@@ -40,7 +40,7 @@ def initialize(client:, api_key:, version: DEFAULT_VERSION)
       #   data = {
       #     customer_code: '1234',
       #     bank_account_type: 'CA',
-      #     account_holder: 'All-Maudra Mayrin',
+      #     bank_account_holder: 'All-Maudra Mayrin',
```
cassiascheffer, line 43: Updated docs to show that this needs the `bank_` prefix.

lib/bambora/rest/www_form_client.rb
```diff
@@ -19,8 +19,8 @@ def post(path:, body:)
         # did not work in testing so I am manually transforming the hash into query parameters here.
         parse_response_body(
           super(
-            path: "#{path}?#{Bambora::Builders::WWWFormParameters.new(body: body)}",
-            body: nil,
+            path: path,
+            body: Bambora::Builders::WWWFormParameters.new(body: body).to_s,
```
cassiascheffer, line 23: Moved this to the body as per the example in the bambora docs. Although, it works in both places just fine.

cassiascheffer (commented):
ᕦ( ✿ ⊙ ͜ʖ ⊙ ✿ )━☆ﾟ.*･｡ﾟ

lib/bambora/rest/client.rb
```diff
@@ -29,6 +29,7 @@ def get(path:, params:, headers:)
       end
 
       def post(path:, body:, headers:)
+        require 'pry'; binding.pry
```
harrylewis: Should this be removed?

harrylewis (changes requested):
Changes look good, and it sounds like you have tested it all with our tests accounts. Marking as "request changes" because of `binding.pry` that is still there. I'll approve once that's removed.

lib/bambora/client.rb
```diff
@@ -154,7 +155,7 @@ def json_client
 
     def www_form_client
       @www_form_client ||= Bambora::Rest::WWWFormClient.new(
-        base_url: base_url,
+        base_url: scripts_api_base_url,
```
cassiascheffer, line 158: Added a second URL in here since the legacy stuff has a different base url.

harrylewis (approved):
🚀

PR #19, opened 2019-12-12
Adding Batch Payments Resource

cassiascheffer:
This PR adds the Batch Payments endpoint to this gem. It allows us to upload a CSV for batch processing.

This has been tested against a test account.

## What to test
- [x] Can submit test batch payments to the Bambora API and get the
expected response back.

### DIFF
--- a/bambora-client.gemspec
+++ b/bambora-client.gemspec
@@ -42,6 +42,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'excon', '< 1.0'
   spec.add_dependency 'faraday', '< 1.0'
   spec.add_dependency 'gyoku', '~> 1.0'
+  spec.add_dependency 'multiparty'
 
   spec.add_development_dependency 'bundler', '~> 1.17'
   spec.add_development_dependency 'pry', '~> 0.12.0'
--- a/lib/bambora/adapters/multipart_mixed_request.rb
+++ b/lib/bambora/adapters/multipart_mixed_request.rb
@@ -5,8 +5,6 @@ module Bambora
     ##
     # Creates headers and a body for a multipart/mixed request with a file and a JSON body.
     class MultipartMixedRequest
-      extend Forwardable
-
       attr_reader :multiparty
 
       def initialize(options = {})
--- a/lib/bambora/adapters/multipart_mixed_request.rb
+++ b/lib/bambora/adapters/multipart_mixed_request.rb
@@ -17,7 +17,9 @@ module Bambora
         multiparty.header.sub(/^Content-Type: /, '').strip
       end
 
-      def_delegator :multiparty, :body, :body
+      def body
+       "#{multiparty.body}\r\n"
+      end
     end
   end
 end
--- a/lib/bambora/adapters/multipart_mixed_request.rb
+++ b/lib/bambora/adapters/multipart_mixed_request.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Bambora
+  module Adapters
+    ##
+    # Creates headers and a body for a multipart/mixed request with a file and a JSON body.
+    class MultipartMixedRequest
+      extend Forwardable
+
+      attr_reader :multiparty
+
+      def initialize(options = {})
+        @multiparty = Multiparty.new { |party| party.parts = options[:multipart_args] }
+      end
+
+      def content_type
+        multiparty.header.sub(/^Content-Type: /, '').strip
+      end
+
+      def_delegator :multiparty, :body, :body
+    end
+  end
+end
--- a/lib/bambora/builders/batch_payment_csv.rb
+++ b/lib/bambora/builders/batch_payment_csv.rb
@@ -18,7 +18,7 @@ module Bambora
         #     account_number: 1223456789,
         #     amount: 10000
         #     reference_nubmer: 1234,
-        #     reccipient_name: 'Hup Podling',
+        #     recipient_name: 'Hup Podling',
         #     customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A',
         #     dynamic_description: 'The Skeksis',
         #   }])
--- a/lib/bambora/builders/batch_payment_csv.rb
+++ b/lib/bambora/builders/batch_payment_csv.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+module Bambora
+  module Builders
+    class BatchPaymentCSV
+      class << self
+        ##
+        # Return a CSV with one transaction per row.
+        #
+        # @see https://dev.na.bambora.com/docs/references/batch_payment/
+        #
+        # @example
+        #   Bambora::Builders::BatchPaymentCSV.build([{
+        #     super_type: 'E',
+        #     transaction_type: 'D',
+        #     institution_number: 12345,
+        #     transit_number: 123,
+        #     account_number: 1223456789,
+        #     amount: 10000
+        #     reference_nubmer: 1234,
+        #     reccipient_name: 'Hup Podling',
+        #     customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A',
+        #     dynamic_description: 'The Skeksis',
+        #   }])
+        #
+        #   # => "E,D,12345,123,123456789,10000,1234,Hup Podling,02355E2e58Bf488EAB4EaFAD7083dB6A,The Skeksis\n"
+        #
+        # @param transactions [Array] an array of transaciton hashes.
+        #
+        # @return [String], a CSV as a string
+        def build(transactions)
+          CSV.generate do |csv|
+            transactions.each do |transaction|
+              csv << transaction.values
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/bambora/client.rb
+++ b/lib/bambora/client.rb
@@ -9,7 +9,7 @@ require 'json'
 # Gems
 require 'faraday' # HTTP Wraper
 require 'gyoku' # XML Builder
-require 'multiparty' # Multipart/mixed requestss
+require 'multiparty' # Multipart/mixed requests
 
 require 'bambora/client/version'
 
--- a/lib/bambora/client.rb
+++ b/lib/bambora/client.rb
@@ -8,12 +8,14 @@ require 'json'
 # Gems
 require 'faraday' # HTTP Wraper
 require 'gyoku' # XML Builder
+require 'multiparty' # Multipart/mixed requestss
 
 require 'bambora/client/version'
 
 # Adapters
 require 'bambora/adapters/response'
 require 'bambora/adapters/json_response'
+require 'bambora/adapters/multipart_mixed_request'
 require 'bambora/adapters/query_string_response'
 require 'bambora/bank/adapters/payment_profile_response'
 
--- a/lib/bambora/client.rb
+++ b/lib/bambora/client.rb
@@ -3,6 +3,7 @@
 # Standard Libraries
 require 'base64'
 require 'cgi'
+require 'csv'
 require 'json'
 
 # Gems
@@ -23,6 +24,7 @@ require 'bambora/bank/adapters/payment_profile_response'
 require 'bambora/builders/headers'
 require 'bambora/builders/xml_request_body'
 require 'bambora/builders/www_form_parameters'
+require 'bambora/builders/batch_payment_csv'
 require 'bambora/bank/builders/payment_profile_params'
 
 # Factories
--- a/lib/bambora/client.rb
+++ b/lib/bambora/client.rb
@@ -174,19 +174,9 @@ module Bambora
     def batch_payment_file_upload_client
       Bambora::Rest::BatchPaymentFileUploadClient.new(
         base_url: base_url,
-        connection: connection,
         merchant_id: merchant_id,
         sub_merchant_id: sub_merchant_id,
       )
-    end
-
-    # TODO: Refactor other Clients to receive a connection as well.
-    def connection
-      @connection ||= Faraday.new(url: base_url) do |faraday|
-        faraday.request :multipart
-        faraday.request :url_encoded
-        faraday.adapter :excon
-      end
     end
   end
 end
--- a/lib/bambora/client.rb
+++ b/lib/bambora/client.rb
@@ -30,6 +30,7 @@ require 'bambora/factories/response_adapter_factory'
 
 # Clients
 require 'bambora/rest/client'
+require 'bambora/rest/batch_payment_file_upload_client'
 require 'bambora/rest/json_client'
 require 'bambora/rest/www_form_client'
 require 'bambora/rest/xml_client'
--- a/lib/bambora/client.rb
+++ b/lib/bambora/client.rb
@@ -146,7 +146,12 @@ module Bambora
 
     def batch_payment_reports; end
 
-    def batch_payments; end
+    def batch_payments(api_key:)
+      @batch_payments ||= Bambora::V1::BatchPaymentResource.new(
+        client: batch_payment_file_upload_client,
+        api_key: api_key
+      )
+    end
 
     private
 
@@ -164,6 +169,24 @@ module Bambora
         merchant_id: merchant_id,
         sub_merchant_id: sub_merchant_id,
       )
+    end
+
+    def batch_payment_file_upload_client
+      Bambora::Rest::BatchPaymentFileUploadClient.new(
+        base_url: base_url,
+        connection: connection,
+        merchant_id: merchant_id,
+        sub_merchant_id: sub_merchant_id,
+      )
+    end
+
+    # TODO: Refactor other Clients to receive a connection as well.
+    def connection
+      @connection ||= Faraday.new(url: base_url) do |faraday|
+        faraday.request :multipart
+        faraday.request :url_encoded
+        faraday.adapter :excon
+      end
     end
   end
 end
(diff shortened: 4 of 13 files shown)


cassiascheffer:
[Bambora Client: Beanstream::BankAPI.new.batch_payments [3]](https://trello.com/c/vFyb0jj2/1574-bambora-client-beanstreambankapinewbatchpayments-3)

lib/bambora/client.rb
```diff
@@ -3,31 +3,36 @@
 # Standard Libraries
 require 'base64'
 require 'cgi'
+require 'csv'
 require 'json'
 
 # Gems
 require 'faraday' # HTTP Wraper
 require 'gyoku' # XML Builder
+require 'multiparty' # Multipart/mixed requestss
```
erawk: Typo: extra "s" at the end.

```suggestion
require 'multiparty' # Multipart/mixed requests
```

bambora-client.gemspec
```diff
@@ -42,6 +42,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'excon', '< 1.0'
   spec.add_dependency 'faraday', '< 1.0'
   spec.add_dependency 'gyoku', '~> 1.0'
+  spec.add_dependency 'multiparty'
```
erawk, line 45: 🎉

lib/bambora/client.rb
bilogub: `requestss`

lib/bambora/v1/batch_payment_resource.rb
```diff
@@ -2,14 +2,48 @@
...
+      #     }],
+      #   )
+      #
+      # @param transactions [Array] of hashes with payment data.
+      # @param opts[:process_now] [Integer] optional. Defaults to 1.
+      # @param opts[:process_date] [String] optional. Must also set process_now to 0.
+      #
+      # @see https://dev.na.bambora.com/docs/references/batch_payment/
+      # @see https://dev.na.bambora.com/docs/guides/batch_payment/
+      def create(transactions, opts = { process_now: 1 })
```
erawk, line 42: If the user includes their own `opts`, and `process_now` isn't included, should it be defaulted to `process_now: 1`? Or should we assume they have intentionally left it out?

erawk (commented):
Thanks for the helpful docs and tests to clarify intent.

lib/bambora/client.rb
```diff
@@ -160,5 +170,13 @@ def www_form_client
         sub_merchant_id: sub_merchant_id,
       )
     end
+
+    def batch_payment_file_upload_client
+      Bambora::Rest::BatchPaymentFileUploadClient.new(
```
bilogub: I see other clients have been memoized, this one is not, method doesn't have arguments to make a different instance anyway.

lib/bambora/rest/batch_payment_file_upload_client.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module Bambora
+  module Rest
+    class BatchPaymentFileUploadClient < Bambora::Rest::Client
+      CONTENT_DISPOSITION = 'form-data'
+      FILE_CONTENT_TYPE = 'text/plain'
```
bilogub: `FILE_CONTENT_TYPE` and `JSON_CONTENT_TYPE` naming is a bit not in agreement and would map use fit a bit better here?

lib/bambora/client.rb
cassiascheffer: 🐍

lib/bambora/rest/batch_payment_file_upload_client.rb
```diff
@@ -0,0 +1,56 @@
...
+          ),
+        ).to_h
+      end
+
+      private
+
+      def init_payload(file_contents, options)
+        return @payload if @payload
+
+        @payload = Bambora::Adapters::MultipartMixedRequest.new(
```
bilogub, line 32: is it memoization in a different way?

lib/bambora/v1/batch_payment_resource.rb
cassiascheffer: I should clarify ln 36. My understanding is that `process_now` overrides `process_date` but if `process_now` is absent and `process_date` is present, `process_now` is assumed to be `false`.

So, yes, we can assume they intentionally left it out. I'll update ln 36 to read `Must exclude process_now or pass { process_now: 0 }`.

lib/bambora/v1/batch_payment_resource.rb
```diff
@@ -2,14 +2,48 @@
...
+      #       reference_nubmer: 1234,
+      #       reccipient_name: 'Hup Podling',
+      #       customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A',
+      #       dynamic_description: 'The Skeksis',
+      #     }],
+      #   )
+      #
+      # @param transactions [Array] of hashes with payment data.
+      # @param opts[:process_now] [Integer] optional. Defaults to 1.
+      # @param opts[:process_date] [String] optional. Must also set process_now to 0.
```
cassiascheffer:
```suggestion
      # @param opts[:process_date] [String] optional. Must exclude +process_now+ or pass +{ process_now: 0 }+
```

lib/bambora/rest/client.rb
```diff
@@ -40,6 +40,8 @@ def delete(path:, headers:)
 
       def connection
         @connection ||= Faraday.new(url: base_url) do |faraday|
+          faraday.request :multipart
+          faraday.request :url_encoded
```
bilogub, line 44: spent some time figuring out this defaults all request to this content-type :) seems those 2 are fallback values.

lib/bambora/rest/batch_payment_file_upload_client.rb
cassiascheffer: Yeah, Rubocop gave me an error because it expects memoized methods to have the same name as the instance variable. But it seemed more semantic to me to name this method `init_` since the instance variable is used elsewhere and it doesn't make sense to use a method that accepts args as an accessor.

cassidycodes: Yeah, Rubocop gave me an error because it expects memoized methods to have the same name as the instance variable. But it seemed more semantic to me to name this method `init_` since the instance variable is used elsewhere and it doesn't make sense to use a method that accepts args as an accessor.

cassidycodes: Yeah, Rubocop gave me an error because it expects memoized methods to have the same name as the instance variable. But it seemed more semantic to me to name this method `init_` since the instance variable is used elsewhere and it doesn't make sense to use a method that accepts args as an accessor.

cassidycodes: Yeah, Rubocop gave me an error because it expects memoized methods to have the same name as the instance variable. But it seemed more semantic to me to name this method `init_` since the instance variable is used elsewhere and it doesn't make sense to use a method that accepts args as an accessor.

lib/bambora/rest/client.rb
cassiascheffer: Ah, yeah, this isn't immediately clear in Faraday, but the middleware stack matches mime types and only executes if there's a match.

https://github.com/lostisland/faraday/blob/9865ebbf8ffdbb2f12bb3af7f02baa10070e8c5a/lib/faraday/request/url_encoded.rb#L18

lib/bambora/builders/batch_payment_csv.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Bambora
+  module Builders
+    class BatchPaymentCSV
+      class << self
+        ##
+        # Return a CSV with one transaction per row.
+        #
+        # @see https://dev.na.bambora.com/docs/references/batch_payment/
```
harrylewis, line 10: Thanks for linking to this documentation!

lib/bambora/builders/batch_payment_csv.rb
```diff
@@ -0,0 +1,41 @@
...
+        #     transaction_type: 'D',
+        #     institution_number: 12345,
+        #     transit_number: 123,
+        #     account_number: 1223456789,
+        #     amount: 10000
+        #     reference_nubmer: 1234,
+        #     reccipient_name: 'Hup Podling',
+        #     customer_code: '02355E2e58Bf488EAB4EaFAD7083dB6A',
+        #     dynamic_description: 'The Skeksis',
+        #   }])
```
harrylewis: The CSV rows must be structured with the data in the correct order. Does Ruby guarantee a certain hash ordering, when reading out the values?

lib/bambora/builders/batch_payment_csv.rb
```diff
@@ -0,0 +1,41 @@
...
+        #
+        #   # => "E,D,12345,123,123456789,10000,1234,Hup Podling,02355E2e58Bf488EAB4EaFAD7083dB6A,The Skeksis\n"
+        #
+        # @param transactions [Array] an array of transaciton hashes.
+        #
+        # @return [String], a CSV as a string
+        def build(transactions)
+          CSV.generate do |csv|
+            transactions.each do |transaction|
+              csv << transaction.values
```
harrylewis, line 34: One way to ensure ordering is by explicitly dereferencing all values in this method, in their correct order. This would be easy given the tests you have. I would advocate for this, because the documentation does not make reference to the fact that the hash must be specified in the order specified above. We could make note of this in the documentation, but that could create confusing issues for the client consumer, when it's passed in incorrectly.

lib/bambora/builders/batch_payment_csv.rb
```diff
@@ -0,0 +1,41 @@
...
+        # @example
+        #   Bambora::Builders::BatchPaymentCSV.build([{
+        #     super_type: 'E',
+        #     transaction_type: 'D',
+        #     institution_number: 12345,
+        #     transit_number: 123,
+        #     account_number: 1223456789,
+        #     amount: 10000
+        #     reference_nubmer: 1234,
+        #     reccipient_name: 'Hup Podling',
```
harrylewis: Small typo.

```suggestion
        #     recipient_name: 'Hup Podling',
```

lib/bambora/rest/batch_payment_file_upload_client.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+module Bambora
+  module Rest
+    class BatchPaymentFileUploadClient < Bambora::Rest::Client
+      CONTENT_DISPOSITION = 'form-data'
+      CONTENT_TYPES ={
```
bilogub: not critical but space for an amused guy `={` w/ mustache is missing :)

bilogub (approved):

lib/bambora/rest/batch_payment_file_upload_client.rb
cassiascheffer: 🙃

bilogub (approved):

lib/bambora/builders/batch_payment_csv.rb
harrylewis: That being said, one test to see this failure would be to scramble the order and pass the data in.

lib/bambora/v1/batch_payment_resource.rb
```diff
@@ -2,14 +2,48 @@
...
+      #     }],
+      #   )
+      #
+      # @param transactions [Array] of hashes with payment data.
+      # @param opts[:process_now] [Integer] optional. Defaults to 1.
+      # @param opts[:process_date] [String] optional. Must exclude +process_now+ or pass +{ process_now: 0 }+
+      #
+      # @see https://dev.na.bambora.com/docs/references/batch_payment/
+      # @see https://dev.na.bambora.com/docs/guides/batch_payment/
+      def create(transactions, opts = { process_now: 1 })
```
harrylewis, line 42: It doesn't seem like we maintain any clear convention for using named parameters and nameless parameters. Based on current resource implementations, does it make sense to make `transactions` a named, keyword parameter?

lib/bambora/rest/batch_payment_file_upload_client.rb
```diff
@@ -0,0 +1,58 @@
...
+
+module Bambora
+  module Rest
+    class BatchPaymentFileUploadClient < Bambora::Rest::Client
+      CONTENT_DISPOSITION = 'form-data'
+      CONTENT_TYPES = {
+        text_plain: 'text/plain',
+        application_json: 'application/json',
+      }.freeze
+      FILE_TYPE_HEADER = { 'FileType' => 'STD' }
```
harrylewis: `#freeze` this as well?

lib/bambora/adapters/multipart_mixed_request.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module Bambora
+  module Adapters
+    ##
+    # Creates headers and a body for a multipart/mixed request with a file and a JSON body.
+    class MultipartMixedRequest
+      attr_reader :multiparty
+
+      def initialize(options = {})
+        @multiparty = Multiparty.new { |party| party.parts = options[:multipart_args] }
```
harrylewis, line 11: Seems like there is a lot of heavy lifting going on in this `Multiparty` functionality 💪.

spec/bambora/adapters/multipart_mixed_request_spec.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+module Bambora
+  module Adapters
```
harrylewis, line 6: This all looks like 💯 test coverage to me!

spec/bambora/adapters/multipart_mixed_request_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        it 'returns the correct content type' do
+          expect(multipart_request.content_type.sub(/-\d+/, '')).to(
+            eq('multipart/form-data; boundary=multiparty-boundary'),
+          )
+        end
+      end
+
+      describe '#body' do
+        it 'returns the body with a carrage return at the end' do
+          expect(multipart_request.body.chars.last(2)).to eq ["\r", "\n"]
```
harrylewis, line 43: How is the body formatted in this case? Does the ordering of the parts matter for Bambora?

spec/bambora/rest/batch_payment_file_upload_client_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+          Content-Disposition: form-data; name="data"; filename="merchant_2.txt"
+          Content-Type: text/plain
+          Content-Transfer-Encoding: binary
+
+          E,C,001,99001,09400313371,10000,1000070001,ACME Corp
+          E,C,002,99002,09400313372,20000,1000070002,John Doe
+          E,C,003,99003,09400313373,30000,1000070003,Jane Doe
+
+          --multiparty-boundary-50428029--
+        BODY
```
harrylewis, line 53: Thank you for putting this in the code. As far as I know, I have only seen this in Bambora's example request in the Postman requests. Great to have this here too.

spec/bambora/v1/batch_payment_resource_spec.rb
```diff
@@ -10,37 +10,54 @@ module V1
...
       end
 
-      subject { described_class.new(client: client, sub_path: '/v1/batch_payments') }
+      subject(:resource) { described_class.new(client: client, api_key: api_key) }
 
-      pending
+      describe '#create' do
+        before do
+        end
+
```
harrylewis: Remove these lines?

erawk (approved):

harrylewis (changes requested):
Wow, thank you for all the heavy lifting you have done in putting this together. It's well broken down, and it's very clear what is going on. I am requesting changes here because of the way the builder puts together the CSV. I think there is potential for error if passing the parameters through in an order that is unexpected by the consumer, without clear documentation as to why that might be the case.

lib/bambora/builders/batch_payment_csv.rb
cassiascheffer: Yes, according to the docs:

>Hashes enumerate their values in the order that the corresponding keys were inserted.

https://devdocs.io/ruby~2.3/hash

lib/bambora/builders/batch_payment_csv.rb
cassiascheffer: As per the docs, each payment type has a different set of keys. Explicitly enumerating over keys that we use would constrain future users until we added keys for all payment types. 

I will add a line to the docs that the hash keys must be ordered the same way the CSV columns are ordered.

cassiascheffer: The docs for reference: https://dev.na.bambora.com/docs/references/batch_payment/

lib/bambora/v1/batch_payment_resource.rb
cassiascheffer: Key word arguments don't work when combined with an options hash because Ruby's underlying keyword implementation relies on a hash itself. The parser would then see `create(transactions: transactions, process_now: 1) as one hash, rather than two.

This style of `arg, options` is fairly common in Ruby, so even though we haven't done it yet in this project, I think it matches a lot of what you see in other places in Ruby land.

spec/bambora/adapters/multipart_mixed_request_spec.rb
cassiascheffer: The body here is a multi-part body as per the specs for the batch_payment_file_upload client.

From my understanding, order doesn't matter because the headers of each section indicate the section's content.

I am only testing the addition of the carriage return here because the body contains a random string as the `multi-part-boundary`.

harrylewis (approved):
Great work, this is some pretty complex stuff!

PR #21, opened 2019-12-17
Bumping Supported Ruby to 2.4.x

cassiascheffer:
This PR bumps the required Ruby Version to >= 2.4.x


erawk (approved):

cassiascheffer:
@erawk Do you think this is too restrictive for others who might use this gem? I could bump to a 2.4 version instead.

erawk:
> @erawk Do you think this is too restrictive for others who might use this gem? I could bump to a 2.4 version instead.

You could keep to the `2.4.x` and run CI tests against latest stable `2.4.x`, `2.5.x`, `2.6.x` rubies too. This could identify incompatible versions of rubies for users of this client early.

cassiascheffer:
@erawk I've updated thi to test against three different rubies.

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - run: |
+          gem install bundler
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
+      - name: Run Tests
+        run: |
+          bundle exec rake spec
```
harrylewis, line 34: We could add linting here too soon, I think most of the code would pass.

erawk (approved):
👋 CircleCI!

👏 For 💎 💎 💎 !

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,34 @@
...
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby: [ '2.4.x', '2.5.x', '2.6.x']
```
harrylewis, line 11: Awesome to have this tested against future versions!

harrylewis (approved):
Great to become more modern, and also to future-proof our work with testing against other rubies.

PR #23, opened 2020-01-03
Batch Report Resource

cassiascheffer:
This PR adds an endpoint for requesting batch reports. The interface follows the patterns of other resources here and remains fairly coarse. We can write a wrapper for this to suit our needs in our own repo.

I have tested this against the test center in staging and got a successful response back.


bambora-client.gemspec
```diff
@@ -42,14 +42,14 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'excon', '< 1.0'
   spec.add_dependency 'faraday', '< 1.0'
   spec.add_dependency 'gyoku', '~> 1.0'
-  spec.add_dependency 'multiparty'
+  spec.add_dependency 'multiparty', '~> 0'
 
-  spec.add_development_dependency 'bundler', '~> 1.17'
+  spec.add_development_dependency 'bundler', '~> 2'
```
cassiascheffer, line 47: Relaxing some restrictions here.

lib/bambora/bank/batch_report_resource.rb
```diff
@@ -0,0 +1,68 @@
...
+        @api_key = api_key
+        @sub_path = '/scripts/reporting/report.aspx'
+      end
+
+      ##
+      # Create a Bank Payment Profile
+      #
+      # @example
+      #   data = {
+      #     rpt_filter_by_1: 'batch_id',
```
cassiascheffer, line 42: Like other resources in this gem, the params passed in are still fairly raw. We'll write a wrapper for this on our end so we can use params that are familiar to us.

lib/bambora/factories/response_adapter_factory.rb
```diff
@@ -13,7 +13,7 @@ def for(response)
           when 'text/html'
             # Currently, the only endpoint that responds wit text/html is /scripts/payment_profiles.asp
             Bambora::Bank::Adapters::PaymentProfileResponse.new(response)
-          else raise Bambora::Client::Error, "Unknown Content Type: #{content_type}"
+          else raise Bambora::Client::Error, "Unknown Content Type: #{content_type}. Response Body: #{response.body}"
```
cassiascheffer, line 16: Found this helpful when I got XML back as an error message.

cassiascheffer (commented):
🗒 🖊

lib/bambora/bank/batch_report_resource.rb
```diff
@@ -0,0 +1,68 @@
...
+
+module Bambora
+  module Bank
+    ##
+    # For making requests to the /scripts/reporting/report.aspx endpoint
+    #
+    # @see https://dev.na.bambora.com/docs/guides/batch_payment/report/
+    class BatchReportResource
+      DEFAULT_REQUEST_PARAMS = {
+        rpt_format: 'JSON',
```
erawk, line 11: Does this report always return JSON, even if you use the XML client to retrieve it? Is this to separate the return type from the client?

erawk (approved):
Nice tests. 

Appreciating how the documentation and test lines well outnumber the client lines of code :)

lib/bambora/bank/batch_report_resource.rb
cassiascheffer: I discovered that it will return XML if you encounter an error. But a success response will always return JSON. That is correct, that this is meant to separate the return type from the client. 

The client inspects the response's `Content-Type` to decide how to parse the response.

I didn't feel it was necessary to add an XML parser to handle the case where we get an error response back since we raise an error and print the response body in the message.

spec/bambora/factories/response_adapter_factory_spec.rb
```diff
@@ -5,9 +5,12 @@
...
   describe ResponseAdapterFactory do
     describe '.for' do
-      let(:response) { instance_double('Faraday::Response', headers: { 'Content-Type' => content_type }) }
+      let(:body) { 'When single shines the triple sun / What was sundered and undone' }
+      let(:response) do
+        instance_double('Faraday::Response', headers: { 'Content-Type' => content_type }, body: body) 
+      end
 
-      subject { described_class.for(response) }
+      subject(:parsed_response) { described_class.for(response) }
```
harrylewis, line 13: This reminds me, should we incorporate `rubocop-rspec` into this repo?

lib/bambora/client.rb
```diff
@@ -90,7 +90,7 @@ def profiles(api_key:)
     # Retrieve a client to make requests to the Payments endpoints.
     #
     # @example
-    #   payments = client.profiles
+    #   payments = client.profiles(api_key: <yourapikey>)
```
harrylewis, line 93: These small changes push us to a great place in terms of automatically generating documentation for this gem. We should look into using `rdoc` or a similar tool for that at some point. And maybe put only key examples into the README. It seems that file is becoming slightly out-of-date, because up to this point it has been a duplication of these comments.

lib/bambora/client.rb
```diff
@@ -144,12 +144,36 @@ def bank_profiles(api_key:)
...
+    #
+    # @param api_key [String] API key for the bank profiles endpoint.
+    #
+    # @return [Bambora::Bank::PaymentProfileResource]
+    def batch_reports(api_key:)
+      @batch_reports = Bambora::Bank::BatchReportResource.new(
+        client: xml_client,
+        api_key: api_key,
+      )
+    end
```
harrylewis, line 171: For consistency, do you want to add specs for this method in `spec/bambora/client_spec.rb`?

lib/bambora/bank/batch_report_resource.rb
```diff
@@ -0,0 +1,68 @@
...
+      #     rpt_operation_type_1: 'EQ',
+      #     rpt_from_date_time: '2019-12-18 00:00:00',
+      #     rpt_to_date_time: '2019-12-18 23:59:59',
+      #     service_name: 'BatchPaymentsEFT',
+      #   }
+      #
+      #  payment_profile_resource.show(data)
+      #
+      #  @params profile_data [Hash] with values as noted in the example.
+      def show(profile_data)
```
harrylewis: Since the method parameters and implementation are generic, functionally I don't think it matters, but the method documentation and the method parameter are named based on payment profile terminology, as opposed to batch report terminology. I just wanted to check if this was intended or not.

lib/bambora/builders/xml_request_body.rb
```diff
@@ -5,15 +5,13 @@ module Builders
...
+      attr_reader :body
 
-      def initialize(body:, response_format: nil)
+      def initialize(body:)
         @body = body
-        @response_format = response_format
       end
 
       def to_s
-        body.merge!(rpt_format: response_format) if response_format
```
harrylewis, line 16: Just curious, why was this needed originally? Was it because we thought we may need to use different return formats, but after going through this, it was realized that we simply want JSON back?

harrylewis (approved):
Functionally everything looks great! I had some comments here that are not necessarily blocking, and could be tackled in some follow-up. Thanks for doing this, we're almost there 🚢

lib/bambora/bank/batch_report_resource.rb
cassiascheffer: Ah, you mean `profile_data`? Good ol' copy-paste. I'll fix that.

lib/bambora/builders/xml_request_body.rb
cassiascheffer: I moved it because it wasn't specific to formatting an XML body but should instead be passed into this class.

lib/bambora/client.rb
cassiascheffer: Done!

harrylewis (approved):
💥

PR #24, opened 2020-01-03
Housekeeping

cassiascheffer:
- Fixes all rubocop errors
- Runs rubocop on every push and comments on the code. You can see this in 376d469

image


bilogub (approved):

harrylewis (approved):
Love that GitHub provides access to all these tools!

PR #28, opened 2020-01-09
Adding CRLF line endings to CSV

cassiascheffer:
According to Bambora, their system runs on windows and needs CRLF line endings rather than CR.


cassiascheffer:
[Bambora::Client CSVs need to end lines with CLRF not CR [2]](https://trello.com/c/TsoTCLoN/1818-bamboraclient-csvs-need-to-end-lines-with-clrf-not-cr-2)

bilogub (approved):

PR #29, opened 2020-01-20
Adding human-readable messages to bank reports

cassiascheffer:
This PR maps "message IDs" on to messages from bambora so that we can use human-readable messages when recording transaction statuses.

Resolves ENG-65


lib/bambora/bank/batch_report_resource.rb
```diff
@@ -51,11 +53,20 @@ def initialize(client:, api_key:)
...
-        client.post(path: sub_path, body: batch_report_body(report_data))
+        add_messages_to_response(
+          client.post(path: sub_path, body: batch_report_body(report_data)),
+        )
       end
 
       private
 
+      def add_messages_to_response(response)
+        response.dig(:response, :record).map! do |record|
```
sweetyclem, line 64: Oh wow I just found out this `dig` method existed, it's amazing!

sweetyclem (approved):
😀

safdarzeeshan (approved):
Nice. One less call to the old beanstream gem :)

From a testing perspective is BatchReportResource being used in the HiMama code (or is that still using BeanStream)?

lib/bambora/bank/batch_report_resource.rb
```diff
@@ -51,11 +53,20 @@ def initialize(client:, api_key:)
...
+          record.merge!(messages: record[:messageId].split(',').map { |id| MESSAGES[id] })
```
safdarzeeshan, line 65: Do you think this would be more intuitive if we name this `error_messages` instead of `messages`?

cassiascheffer: The field from Bambora is `messageId` and not all of the messages are errors, so I think it's better to keep the field named `message`.

`BatchReportResource` is used in HiMama indirectly by calling `Bambora::Client.new(args).batch_reports(api_key)`

PR #34, opened 2020-03-04
Create `gempush.yml` to use an action to publish to Github and Rubygems

erawk:
Currently, to publish a new version of the gem, we have to do this manually from a development machine. This automates that process.

### DIFF
--- a/.github/workflows/gempush.yml
+++ b/.github/workflows/gempush.yml
@@ -1,0 +1,44 @@
+name: Ruby Gem
+
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Build + Publish
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby 2.6
+      uses: actions/setup-ruby@v1
+      with:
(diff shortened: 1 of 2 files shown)


erawk:
This should work, but we need to run the action only when we create a new tag.

.github/workflows/gempush.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby 2.6
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: 2.6.x
+
+    - name: Publish to RubyGems
```
erawk, line 20: Our `gemspec` only allows pushing to Rubygems, so only run this one pipeline.

.github/workflows/gempush.yml
```diff
@@ -0,0 +1,31 @@
+name: Ruby Gem
+
+on:
+  push:
+    tags:
+      - v* # Only run when we carve a new tag
```
cassiascheffer, line 6: 🙏

cassiascheffer (approved):
Nice! Thank you!

harrylewis (approved):
🙌

PR #35, opened 2020-04-13
Bugfix: Handle null record from Bambora Batch Bank Report queries

erawk:
**Description**

For some Bambora Batch Bank Report queries, the `report` is null. Previously, we were attempting to call `map!` on a `nil` object. This avoids the call and returns the undecorated `response` back up the stack. 

The test included demonstrates the bug, and the fix makes it pass.

This is related to the HiMama ticket: https://himama.atlassian.net/browse/ENG-505
And the Sentry bug: https://sentry.io/organizations/himama-inc/issues/1520181007/

### DIFF
--- a/lib/bambora/bank/batch_report_resource.rb
+++ b/lib/bambora/bank/batch_report_resource.rb
@@ -61,9 +61,14 @@ module Bambora
       private
 
       def add_messages_to_response(response)
+        # bambora can return null or empty record results
+        records = response.dig(:response, :record)
+        return response if records.nil?
+
         response.dig(:response, :record).map! do |record|
           record.merge!(messages: record[:messageId].split(',').map { |id| MESSAGES[id] })
         end
+
         response
       end
 
--- a/lib/bambora/bank/batch_report_resource.rb
+++ b/lib/bambora/bank/batch_report_resource.rb
@@ -61,9 +61,8 @@ module Bambora
       private
 
       def add_messages_to_response(response)
-        # bambora can return null or empty record results
-        records = response.dig(:response, :record)
-        return response if records.nil?
+        # bambora can return null or empty record results, don't decorate with messages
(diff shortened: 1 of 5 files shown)


lib/bambora/bank/batch_report_resource.rb
```diff
@@ -61,9 +61,14 @@ def show(report_data)
       private
 
       def add_messages_to_response(response)
+        # bambora can return null or empty record results
+        records = response.dig(:response, :record)
+        return response if records.nil?
```
cassiascheffer: Minor naming thing here that is a bit confusing in the response from Bambora. There are two nodes `records` (plural) that has a count of the total number of records and `record` (singular) that contains all the records.

It would be more clear to either do:

```suggestion
        record = response.dig(:response, :record)
        return response if record.nil?
```

or

```suggestion
        total = response.dig(:response, :records, :total)
        return response if total.zero?
```

This way the variable name matches the node name.

cassiascheffer (dismissed):
Approving as this solves the problem, but I left one comment about clarifying the variable name. 

Thanks for doing this!

lib/bambora/bank/batch_report_resource.rb
harrylewis: I was definitely scratching my head over Bambora's API about this. I think a minor change like this could help with that understanding.

harrylewis (dismissed):
That was easy. Good time to version bump too 🚀

harrylewis (approved):

cassiascheffer (approved):

lib/bambora/bank/batch_report_resource.rb
erawk: - [x] Updated this to remove the unneeded variable

Thanks for the feedback. Marking as resolved.

PR #37, opened 2020-05-11
Implement payment profile retrieval method

harrylewis:
**Description**

This PR implements a `#get` method for the `ProfileResource` to retrieve data for a payment profile.

**What To Test**

+ [x] Can retrieve payment profile data

**Deployment Instructions**

+ No action required

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.2.0 (2020-05-11)
+
+- Add `Bambora::V1::ProfileResource#get` method
+
 ## 0.1.3 (2020-04-30)
 
 - Ensure Batch Report returns empty array `:record` in response at the very least
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    bambora-client (0.1.2)
+    bambora-client (0.1.3)
       excon (< 1.0)
       faraday (< 1.0)
       gyoku (~> 1.0)
@@ -29,7 +29,7 @@ GEM
     method_source (0.9.2)
     mime-types (3.3.1)
       mime-types-data (~> 3.2015)
-    mime-types-data (3.2019.1009)
+    mime-types-data (3.2020.0425)
     multipart-post (2.1.1)
     multiparty (0.2.0)
       mime-types
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    bambora-client (0.1.0)
+    bambora-client (0.1.1)
       excon (< 1.0)
       faraday (< 1.0)
       gyoku (~> 1.0)
@@ -19,7 +19,7 @@ GEM
     crack (0.4.3)
       safe_yaml (~> 1.0.0)
     diff-lcs (1.3)
-    excon (0.71.0)
+    excon (0.70.0)
     faraday (0.17.1)
       multipart-post (>= 1.2, < 3)
     gyoku (1.3.1)
@@ -90,4 +90,4 @@ DEPENDENCIES
   webmock (~> 3.7)
 
 BUNDLED WITH
-   2.1.0
+   2.1.1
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.2.0 (2020-05-11)
```
harrylewis, line 1: Since this is a new feature, and not a bug fix, I felt it warranted a minor version update. In keeping with [SemVer](https://semver.org/), this update contains no breaking changes.

harrylewis (commented):
Let's `GET` those profiles!

lib/bambora/client/version.rb
```diff
@@ -2,6 +2,6 @@
 
 module Bambora
   class Client
-    VERSION = '0.1.3'
+    VERSION = '0.2.0'
```
safdarzeeshan, line 5: Why does this get updated?

safdarzeeshan (approved):
Nice to see this `GET` done :)

lib/bambora/client/version.rb
harrylewis: I updated this manually. I am increasing the gem version number to signify new changes are present with this PR. This will also deploy the new version to RubyGems with a GitHub action.

lib/bambora/v1/profile_resource.rb
```diff
@@ -58,6 +58,48 @@ def create(card_data)
...
       end
 
+      ##
+      # Get a Bambora payment profile given a customer code.
+      #
+      # @example
+      #
+      #   client = Bambora::Rest::JSONClient(base_url: '...', api_key: '...', merchant_id: '...')
+      #   profiles = Bambora::V1::ProfileResource(client: client)
+      #   customer_code = '02355E2e58Bf488EAB4EaFAD7083dB6A'
```
erawk, line 68: Is this a real customer code?

Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    bambora-client (0.1.2)
+    bambora-client (0.1.3)
```
erawk, line 4: Strange. Is this always a version behind because the `gem` hasn't been released yet?

erawk (approved):

lib/bambora/v1/profile_resource.rb
harrylewis: It used to be a real API key, not even a customer code 😰. That key was cycled though about a month ago. This no longer represents any live customer code or API key.

Gemfile.lock
harrylewis: This *is* strange. I was wondering why we even are maintaining a `Gemfile.lock` in the codebase. The lock file is meant to specify dependencies for a certain machine/environment.

Since this is a gem, dependency versions should be stated in the `*.gemspec` file, and not in the lock file. [This GitHub issue explains the situation well](https://github.com/radar/guides/issues/14), and references an article from Yehuda Katz (albeit from 2010) explaining that the lock file should **not** be checked in for a gem. This is maintained further on [StackOverflow](https://stackoverflow.com/questions/4151495/should-gemfile-lock-be-included-in-gitignore).

A few years ago, Bundler modified the gem setup to include a lock file, [explaining that in some cases, it is desirable to check this in](https://bundler.io/blog/2018/01/17/making-gem-development-a-little-better.html). I don't think our gem falls into this category. I'm spinning up a PR to remove it.

lib/bambora/v1/profile_resource.rb
```diff
@@ -58,6 +58,48 @@ def create(card_data)
...
+      #
+      #   profiles.get(customer_code: customer_code)
+      #   # => {
+      #   #      :code => 1,
```
cassiascheffer, line 72: Curious, why the hash rocket syntax here if the keys are all symbolified?

harrylewis: @cassidycodes the reason for that is to emulate as close as possible what one would see in the REPL when running this.

PR #40, opened 2020-10-01
update faraday version to 1.0.1

antodoms:
**Description**
* updated faraday to latest version

### DIFF
--- a/bambora-client.gemspec
+++ b/bambora-client.gemspec
@@ -40,7 +40,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.4.6'
 
   spec.add_dependency 'excon', '< 1.0'
-  spec.add_dependency 'faraday', '< 1.0'
+  spec.add_dependency 'faraday', '1.0.1'
   spec.add_dependency 'gyoku', '~> 1.0'
   spec.add_dependency 'multiparty', '~> 0'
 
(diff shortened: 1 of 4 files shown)


harrylewis:
Hey @antodoms! 👋 

Thanks for the contribution. It's good to know `faraday` has been updated, and it would be great to get this change into the gem. Would you consider re-opening this?

TylerRick:
https://github.com/lostisland/faraday is at 2.7.4 now. Why is this gem locked to faraday < 1.0 and can we update to the latest?

And no, this is not actually causing me problems or blocking me at the moment, but it did surprise me when I switched to this gem from a much older beanstream gem and saw the faraday version change drastically:
```
    -faraday (2.7.4)
    +faraday (0.17.6)
```

harrylewis:
Hey @TylerRick, thanks for the comment 👋🏻 

The reason for locking it at that version was because that the main use case for the gem maintainers requires a lower version of `faraday`. We didn't want to mislead folks into thinking it had been tested and evaluated with higher versions of `faraday`.

That being said, I don't any issues with relaxing those version constraints, as long as it is made clear in the `README` that it has only been tested with lower versions of `faraday`.

If you wish, you can also provide more insight into how it is functioning on higher versions of `faraday`. Let me know what you think!

TylerRick:
As a general practice, it seems like gems should remain compatible with current/recent versions of their dependencies. Otherwise, if you have 2 gems in your project that depend on `faraday`, and one of those is locked to faraday `0.17.6`, and the other requires at least `2.x`, then you have a problem...

This is currently only a hypothetical problem for us, since this is the only gem in our project that uses `faraday`, but just wanted to chime in with a "we should update someday"... :smile:

harrylewis:
@TylerRick the day has come! We have updated the `faraday` requirements to a more modern version.

PR #42, opened 2021-05-31
Fix GitHub actions

harrylewis:
**Description**

This PR fixes the GitHub actions to be run for this repository.

**What To Test**

- [x] CI passes

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -18,8 +18,6 @@ jobs:
           ${{ runner.os }}-gem-
     - name: Setup ruby
       uses: ruby/setup-ruby@v1
-      with:
-        ruby-version: ${{ matrix.ruby }}
     - run: |
         gem install bundler
         bundle config path vendor/bundle
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -7,7 +7,22 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.ref }}
+    - name: Cache Gems
+      uses: actions/cache@v1
+      with:
+        path: vendor/bundle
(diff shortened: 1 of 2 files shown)


.github/workflows/linter.yml
```diff
@@ -7,7 +7,22 @@ jobs:
...
+    - name: Setup ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby }}
+    - run: |
+        gem install bundler
+        bundle config path vendor/bundle
+        bundle install --jobs 4 --retry 3
     - name: Rubocop checks
-      uses: gimenete/rubocop-action@1.0
```
harrylewis, line 11: This prebuilt action only installed the latest version of `rubocop`, and not the one contained in the gemspec. While it is ideal to pull the latest version, we are at a much older version. Due to this, there are a significant number of changes to be made, and in general, it's ideal to have more control over which version is used.

The changes above are copied from the Ruby action below, and facilitate running the RuboCop check in a custom way, using the `rubocop` version specified in the gemspec.

.github/workflows/ruby.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gem-
       - name: Setup ruby
-        uses: actions/setup-ruby@v1
+        uses: ruby/setup-ruby@v1
```
harrylewis, line 25: `actions/setup-ruby@v1` has been deprecated in favour of `ruby/setup-ruby@v1`.

.github/workflows/ruby.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [ '2.4.x', '2.5.x', '2.6.x']
+        ruby: ['2.4', '2.5', '2.6']
```
harrylewis, line 11: `ruby/setup-ruby@v1` requires the removal of the trailing `.x`.

PR #43, opened 2021-05-31
Implement `Bambora::V1::PaymentResource#get`

harrylewis:
**Description**

This PR implements `Bambora::V1::PaymentResource#get`, allowing a client to get a payments' details.

**What To Test**

- [x] CI passes.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -40,9 +40,6 @@ Style/TrailingCommaInHashLiteral:
 Style/TrailingCommaInArguments:
   EnforcedStyleForMultiline: comma
 
-Layout/ParameterAlignment:
-  EnforcedStyle: with_fixed_indentation
-
 Layout/EmptyLineAfterMagicComment:
   Enabled: false
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.0 (2021-05-31)
+
+- Add `Bambora::V1::PaymentResource#get` method
+
 ## 0.3.0 (2020-05-27)
 
 - Add `Bambora::V1::ProfileResource#update` method
--- a/lib/bambora/v1/payment_resource.rb
+++ b/lib/bambora/v1/payment_resource.rb
@@ -87,7 +87,7 @@ module Bambora
       #   payments = Bambora::V1::PaymentResource(client: client)
       #   payments.get(transaction_id: 1000341)
       #
-      # @param transaction_id [Integer] An integeridentifier for the associated transaction.
+      # @param transaction_id [Integer] An integer identifier for the associated transaction.
       #
       # @return [Hash] Transaction details.
       def get(transaction_id:)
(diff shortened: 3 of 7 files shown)


.rubocop.yml
```diff
@@ -40,9 +40,6 @@ Style/TrailingCommaInHashLiteral:
 Style/TrailingCommaInArguments:
   EnforcedStyleForMultiline: comma
 
-Layout/ParameterAlignment:
-  EnforcedStyle: with_fixed_indentation
```
harrylewis, line 44: RuboCop was complaining that this cop no longer exists, so I removed it.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.4.0 (2021-05-31)
+
+- Add `Bambora::V1::PaymentResource#get` method
```
harrylewis, line 3: Cutting a new release, `0.4.0`, to accommodate this additional feature. Once this is merged, I will tag `main` appropriately, triggering a new release to be pushed to RubyGems through the `gempush` GitHub action.

lib/bambora/v1/payment_resource.rb
```diff
@@ -77,6 +77,22 @@ def create_with_payment_profile(customer_code:, amount:, card_id: 1, complete: f
...
+      #   client = Bambora::Rest::JSONClient(base_url: '...', api_key: '...', merchant_id: '...')
+      #   payments = Bambora::V1::PaymentResource(client: client)
+      #   payments.get(transaction_id: 1000341)
+      #
+      # @param transaction_id [Integer] An integer identifier for the associated transaction.
+      #
+      # @return [Hash] Transaction details.
+      def get(transaction_id:)
+        client.get(path: "#{sub_path}/#{transaction_id}", api_key: api_key)
+      end
```
cristianoventura, line 95: @harrylewis Does `client.get` also handle HTTP errors? What happens in case this returns a 401 or 502 for example?

harrylewis: @cristianoventura good question. The `client` is an inherited instance of `Bambora::Rest::Client`, which has a method `#parse_response_body` shown below, which is called for all networking operations.

https://github.com/HiMamaInc/bambora-client/blob/e057af28cc490e3a49269f79206ee71cf4715deb/lib/bambora/rest/client.rb#L62-L64

You can see this calls on an adapter, which eventually delegates to an inherited instance of `Bambora::Response`, where `#error_response` is defined.

https://github.com/HiMamaInc/bambora-client/blob/e057af28cc490e3a49269f79206ee71cf4715deb/lib/bambora/adapters/response.rb#L36-L38

No `raise` is happening. Rather, a response is returned from `Faraday`, the dependency we use for networking, and eventually we will return this error response to the caller to handle.

cristianoventura: Thanks for clarifying! 👍 🙌

cristianoventura (approved):
Looks good to me! 🚀

mercyoseni (approved):
LGTM

PR #46, opened 2023-01-04
Bump Ruby version

trooney-himama:
Bump Ruby version to 2.6, test 2.7


trooney-himama:
@harrylewis This looks safe. How safe is this? 😆

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.4.1 (2023-01-04)
+
+- Update to Ruby 2.6.6
```
harrylewis: Can we specify we are upgrading the "Development" version of Ruby? Additionally, if we are removing support for certain Ruby version with the `gemspec` changes below, in light of not being officially public (version `0.x.x`) can we make this a minor version update, to `0.5.0`, and indicate those removals in here?

Alternatively, if we don't want to go that far, we can just add `2.7` to the testing matrix above (and not also remove `2.4` and `2.5`), and call it a day.

trooney-himama: @harrylewis Updated version to 0.5.0 and CHANGELOG to reflect that. Dropping support for legacy ruby versions.

harrylewis (approved):
Thanks Tyler!

PR #50, opened 2023-12-01
[CM-278] allow retrieve all sub merch

jazziining:
**Description:**
This PR is to allow passing `sub merchant id` when calling the `.show` where we use the `rpt_merchant_id` like all other query elements. This will enable retrieving the report not only for a specific sub merchant but also for `all live` sub merchants.

This update should not affect any current usage, as you should still be able to pass the `sub_merchant_id` when calling the `Bambora::Client.new`. In the mean time, you can also get rid of that and pass additional element when calling the `client.show` where you can specify the sub merchant id by using the key `rpt_merchant_id`

**Jira**: [CM-216](https://himama.atlassian.net/browse/CM-216)

**Test:**
- [ ] I have added the test in the spec
- [ ] Test by adding the following in the console file and run `./bin/console`
```
elements = {
  service_name: 'BatchPaymentsEFT',
  rpt_filter_by_1: 'returned_date',
  rpt_operation_type_1: 'GT',
  rpt_filter_value_1: '2023-08-01',
  rpt_merchant_id: 'AllLive',
}

client ||= begin
  url = ENV.fetch('BAMBORA_SCRIPTS_API_BASE_URL')
  partner_merchant_id = ENV.fetch("BAMBORA_CAD_MERCHANT_ID")
  reporting_api_key = ENV.fetch("BAMBORA_CAD_REPORTING_API_KEY")

  ::Bambora::Client
    .new(scripts_api_base_url: url, merchant_id: partner_merchant_id)
    .batch_reports(api_key: reporting_api_key)
end

response = client.show(elements)

puts response
```
- [ ] Test by adding the following in the console file and run `./bin/console`
```
elements = {
  service_name: 'BatchPaymentsEFT',
  rpt_filter_by_1: 'batch_id',
  rpt_operation_type_1: 'EQ',
  rpt_filter_value_1: 10000004,
}

client ||= begin
  url = ENV.fetch('BAMBORA_SCRIPTS_API_BASE_URL')
  partner_merchant_id = ENV.fetch("BAMBORA_CAD_MERCHANT_ID")
  reporting_api_key = ENV.fetch("BAMBORA_CAD_REPORTING_API_KEY")

  ::Bambora::Client
    .new(scripts_api_base_url: url, merchant_id: partner_merchant_id, sub_merchant_id: sub_merchan_id)
    .batch_reports(api_key: reporting_api_key)
end

response = client.show(elements)

puts response
```


lib/bambora/client.rb
```diff
@@ -163,7 +163,7 @@ def bank_profiles(api_key:)
     #
     # @param api_key [String] API key for the bank profiles endpoint.
     #
-    # @return [Bambora::Bank::PaymentProfileResource]
+    # @return [Bambora::Bank::BatchReportResource]
     def batch_reports(api_key:)
       @batch_reports = Bambora::Bank::BatchReportResource.new(
```
jazziining, line 168: Change here as you can see from the line 168 the expected return is `Bambora::Bank::BatchReportResource` not `Bambora::Bank::PaymentProfileResource`

spec/bambora/bank/batch_report_resource_spec.rb
```diff
@@ -6,12 +6,15 @@ module Bambora
...
       subject(:reports) { described_class.new(client: client, api_key: api_key) }
+      subject(:reports_without_sub_merchant_id) do
+        described_class.new(client: client_withot_sub_merchant_id, api_key: api_key)
+      end
 
       let(:api_key) { 'fakekey' }
       let(:merchant_id) { 1 }
       let(:sub_merchant_id) { 2 }
       let(:base_url) { 'https://sandbox-web.na.bambora.com' }
-      let(:headers) { { 'Authorization' => 'Passcode MTpmYWtla2V5', 'Sub-Merchant-ID' => sub_merchant_id } }
```
jazziining, line 14: Remove this line as it is not used or needed.

lib/bambora/bank/batch_report_resource.rb
```diff
@@ -79,11 +79,14 @@ def add_messages_to_response(response)
...
+        additional_data = {
           merchant_id: client.merchant_id,
           pass_code: api_key,
-          sub_merchant_id: client.sub_merchant_id,
-        )
+        }
+
+        params = DEFAULT_REQUEST_PARAMS.merge(additional_data).merge(request_data)
+
+        client.sub_merchant_id.nil? ? params : params.merge(sub_merchant_id: client.sub_merchant_id)
```
cristianoventura: Not a blocker, but an alternative could be using `.compact` to remove `sub_merchant_id` if it's `nil`.

https://ruby-doc.org/core-2.4.2/Hash.html#method-i-compact

spec/bambora/bank/batch_report_resource_spec.rb
```diff
@@ -96,6 +99,15 @@ module Bank
         )
       end
 
+      let(:client_withot_sub_merchant_id) do
```
cristianoventura:
```suggestion
      let(:client_without_sub_merchant_id) do
```

lib/bambora/bank/batch_report_resource.rb
cristianoventura: Also, does this new behaviour need to be documented in the [Readme](https://github.com/HiMamaInc/bambora-client/blob/main/README.md)?

cristianoventura (dismissed):
Looks good to me!

If we are planning to release this publicly right away I'd recommend bumping the patch or minor [version](https://github.com/HiMamaInc/bambora-client/blob/main/lib/bambora/client/version.rb) as well as updating the [changelog](https://github.com/HiMamaInc/bambora-client/blob/main/CHANGELOG.md). If we do this, we probably need to make a release in [ruby gems](https://rubygems.org/gems/bambora-client) too, but if there are more related changes coming up, we can pack all of them together and release all at once.

lib/bambora/bank/batch_report_resource.rb
harrylewis: The suggestion from @cristianoventura is really good here. It may remove the need for the conditional logic, as long as we change the merge order, and then call `#compact`.

```ruby
additional_data = {
  merchant_id: client.merchant_id,
  pass_code: api_key,
  sub_merchant_id: client.sub_merchant_id,
}

DEFAULT_REQUEST_PARAMS.merge(additional_data).merge(request_data).compact
```

harrylewis: @cristianoventura that's a good idea to update the README, although it needs some work in general. I'll note this for future work to update the README. Is that cool with you?

harrylewis:
@cristianoventura good call on the gem versioning. Since this is a change we are still going to be exercising and evaluating, I suggest we leave the versioning out for now until it has been living in production for some time, and then we can cut a new release.

Does that sound reasonable to you?

harrylewis:
@cristianoventura @jazziining that being said, I think we could update the CHANGELOG, with a new **Unreleased** heading, with a point about this new change. Do you mind doing that Jasmine?

cristianoventura (dismissed):

lib/bambora/bank/batch_report_resource.rb
```diff
@@ -83,7 +83,7 @@ def batch_report_body(request_data)
           merchant_id: client.merchant_id,
           pass_code: api_key,
           sub_merchant_id: client.sub_merchant_id,
-        )
+        ).compact
```
harrylewis, line 86: Oh wow, we can even keep the previous implementation mostly intact with that suggestion 🙏🏻

harrylewis (approved):
Small change, lots of testing, and a much more flexible API! Great work @jazziining!

PR #51, opened 2024-02-01
Introduce new `Bambora::Credentials` class

harrylewis:
# Description

This code change introduces a new class, `Bambora::Credentials`.

The intention of this class is to act as a consolidated store of a merchant's API credentials that can be easily referenced and passed around. This can hold a set of sub-merchant credentials or partner merchant credentials.

For now, this class only supports the following credential items.

- Merchant ID
- Reporting Passcode

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -18,9 +18,7 @@ jobs:
           ${{ runner.os }}-gem-
     - name: Setup ruby
       uses: ruby/setup-ruby@v1
-    - run: |
-        gem install bundler
-        bundle config path vendor/bundle
-        bundle install --jobs 4 --retry 3
+    - name: Install dependencies
+      run: bundle install
     - name: Rubocop checks
       run: bundle exec rubocop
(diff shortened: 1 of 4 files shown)


.github/workflows/linter.yml
```diff
@@ -18,9 +18,7 @@ jobs:
           ${{ runner.os }}-gem-
     - name: Setup ruby
       uses: ruby/setup-ruby@v1
-    - run: |
-        gem install bundler
-        bundle config path vendor/bundle
-        bundle install --jobs 4 --retry 3
+    - name: Install dependencies
+      run: bundle install
```
harrylewis, line 22: This step was failing as you can [see here](https://github.com/HiMamaInc/bambora-client/actions/runs/7735401383/job/21090993234?pr=51). It was failing to install the correct version of `bundler` for the given Ruby version.


The https://github.com/ruby/setup-ruby GitHub Action comes with [Bundler installed by default](https://github.com/ruby/setup-ruby#bundler), so we don't really need to re-install `bundler`. This ensures a correct version of `bundler` is installed, and now the tests and linting pass.

jazziining (approved):
✨ nice addition for future flexibility

